PR #2, opened 2017-01-31
Server API Spec

ClaytonPassmore:
I tried to spec out the API for the server.

This is just what I could come up with for an initial version.
If there is anything that I'm missing, please let me know!

### DIFF
--- a/Spec/API.md
+++ b/Spec/API.md
@@ -1,0 +1,381 @@
+# API
+Server-side API routes.
+- [User Management](#user-management)
+    - [User Creation](#user-creation)
+    - [User Authentication](#user-authentication)
+    - [User Information](#user-information)
+- [Deck Manipulation](#deck-manipulation)
+    - [Deck Creation](#deck-creation)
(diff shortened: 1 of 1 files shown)


ClaytonPassmore:
Tried to add versioning to the API... versioning is going to suck :cry: 

Whenever we make deck edits or card edits, we need to pass the "parent" or "current" version of the deck, so that we know which version the edit is being made on. The idea is that if the version does not match with the latest version on the server, the edit would get revoked.

Let me know your thoughts!

Spec/API.md
```diff
@@ -0,0 +1,330 @@
...
+Request:
+```
+POST /api/deck/<uuid>/rate
+{
+    "rating": <rating>
+}
+```
+Where `<rating>` is an integer in the range \[0, 5\] and `<uuid>` is the unique identifier of a public deck.
+
+## Discoverability
```
mattyu007: What about an endpoint for searching public decks?

ClaytonPassmore: Good call! Added.

PR #3, opened 2017-02-13
Preliminary schema

ClaytonPassmore:
Let me know what you guys think of this schema.

I haven't actually made sure this is syntactically correct yet, I'm just looking to see what comments you guys have about it at a high level.

### DIFF
--- a/Server/SQL/schema.sql
+++ b/Server/SQL/schema.sql
@@ -1,0 +1,74 @@
+CREATE TABLE IF NOT EXISTS User
+(
+    uuid VARCHAR(36) NOT NULL PRIMARY KEY,
+    access_token VARCHAR(36) NOT NULL,
+    name VARCHAR(255) NOT NULL
+);
+
+
+CREATE TABLE IF NOT EXISTS Deck
+(
+    uuid VARCHAR(36) NOT NULL PRIMARY KEY,
+    name VARCHAR(255) NOT NULL,
+    owner VARCHAR(36) NOT NULL,
+    rating FLOAT DEFAULT NULL,
+    num_ratings INTEGER NOT NULL DEFAULT 0,
+    public BOOLEAN NOT NULL DEFAULT FALSE,
+    deleted BOOLEAN NOT NULL DEFAULT FALSE,
+    version INTEGER NOT NULL DEFAULT 0,
+    share_code VARCHAR(8) UNIQUE,
+
+    FOREIGN KEY (owner) REFERENCES User(uuid)
+);
+
+
+CREATE TABLE IF NOT EXISTS Card
+(
+    uuid INTEGER NOT NULL AUTO_INCREMENT PRIMARY KEY,
+    deck_id VARCHAR(36) NOT NULL,
+    front VARCHAR(255) NOT NULL,
+    back VARCHAR(255) NOT NULL,
+    position INTEGER NOT NULL,
+
+    FOREIGN KEY (deck_id) REFERENCES Deck(uuid) ON DELETE CASCADE
+);
+
+
+CREATE TABLE IF NOT EXISTS NeedReview
+(
+    card_id INTEGER NOT NULL,
+    user_id VARCHAR(36) NOT NULL,
+
+    PRIMARY KEY (card_id, user_id),
+    FOREIGN KEY (card_id) REFERENCES Card(uuid) ON DELETE CASCADE,
+    FOREIGN KEY (user_id) REFERENCES User(uuid)
+);
+
+
+CREATE TABLE IF NOT EXISTS Library
+(
+    user_id VARCHAR(36) NOT NULL,
+    deck_id VARCHAR(36) NOT NULL,
+
+    PRIMARY KEY (user_id, deck_id),
+    FOREIGN KEY (user_id) REFERENCES User(uuid),
+    FOREIGN KEY (deck_id) REFERENCES Deck(uuid)
+);
+
+
+CREATE TABLE IF NOT EXISTS Tag
+(
+    uuid INTEGER NOT NULL AUTO_INCREMENT PRIMARY KEY,
+    tag VARCHAR(255) NOT NULL
+);
+
+
+CREATE TABLE IF NOT EXISTS DeckTag
+(
+    tag_id INTEGER NOT NULL,
+    deck_id VARCHAR(36) NOT NULL,
+
+    PRIMARY KEY (tag_id, deck_id),
+    FOREIGN KEY (tag_id) REFERENCES Tag(uuid),
+    FOREIGN KEY (deck_id) REFERENCES Deck(uuid)
+);


Server/SQL/schema.sql
```diff
@@ -0,0 +1,63 @@
+CREATE TABLE IF NOT EXISTS Card
+(
+    id INTEGER NOT NULL AUTO_INCREMENT PRIMARY KEY,
```
mattyu007: Is there any reason this isn't a UUID?

Server/SQL/schema.sql
```diff
@@ -0,0 +1,63 @@
...
+    deck_id VARCHAR(255) NOT NULL,
```
mattyu007: UUID fields can be shrunk down to `varchar(36)` to save space (or even `binary(16)` according to [Storing UUID Values in MySQL Tables](http://mysqlserverteam.com/storing-uuid-values-in-mysql-tables/)).

I'm slightly surprised MySQL doesn't have a native `UUID` type...

Server/SQL/schema.sql
```diff
@@ -0,0 +1,63 @@
...
+    needs_review BOOLEAN NOT NULL DEFAULT FALSE,
+    front VARCHAR(255) NOT NULL,
+    back VARCHAR(255) NOT NULL,
+    position INTEGER NOT NULL
+
+    FOREIGN KEY (deck_id) REFERENCES Deck(uuid)
```
mattyu007: Should we add `ON DELETE CASCADE` to this foreign key? Would ensure that if a parent deck is actually deleted from `Deck`, we don't end up with orphaned `Card` records. If we don't, I believe MySQL will throw an error if we try to delete a non-empty deck, which might be what we want since we don't anticipate deleting decks at all?

Server/SQL/schema.sql
```diff
@@ -0,0 +1,63 @@
...
+);
+
+
+CREATE TABLE IF NOT EXISTS Deck
+(
+    uuid VARCHAR(255) NOT NULL PRIMARY KEY,
+    name VARCHAR(255) NOT NULL,
+    owner VARCHAR(255),  -- theoretically could be null if a user deletes their account?
```
mattyu007: This would depend on whether we actually allow people to delete their accounts, if we allow people to just deactivate their accounts, or if we don't support account deletion at all.

Server/SQL/schema.sql
```diff
@@ -0,0 +1,63 @@
...
+    rating FLOAT DEFAULT NULL,
+    num_raitings INTEGER NOT NULL DEFAULT 0,
```
mattyu007: Typo ;)

Server/SQL/schema.sql
```diff
@@ -0,0 +1,63 @@
...
+    public BOOLEAN NOT NULL DEFAULT FALSE,
```
mattyu007: Just to confirm: We have 3 useful combinations of `public` and `share_code`?

| `public` | `share_code` | Sharing Status                    |
|----------|--------------|-----------------------------------|
| `false`  | Any          | Deck is **private**               |
| `true`   | `NULL`       | Deck is **public**                |
| `true`   | String       | Deck is **shared** using the code |

I wish we could just declare an enum like

```swift
enum SharingStatus {
   case private
   case shared(code: String)
   case public
}
```

😕

Server/SQL/schema.sql
```diff
@@ -0,0 +1,63 @@
...
+    deleted BOOLEAN NOT NULL DEFAULT FALSE,
+    version INTEGER NOT NULL DEFAULT 0,
+    share_code VARCHAR(255) UNIQUE,
```
mattyu007: This could be shrunk down to `varchar(6)` or `varchar(8)` depending on how many characters we use for the sharing code.

Server/SQL/schema.sql
```diff
@@ -0,0 +1,63 @@
...
+CREATE TABLE IF NOT EXISTS Tag
+(
+    id INTEGER NOT NULL AUTO_INCREMENT PRIMARY KEY,
+    tag VARCHAR(255) NOT NULL
+);
+
+
+CREATE TABLE IF NOT EXISTS DeckTag
+(
+    tag_id VARCHAR(255) NOT NULL,
```
mattyu007: Should this be `INTEGER` to match the declaration of `Tag.id`?

Server/SQL/schema.sql
```diff
@@ -0,0 +1,63 @@
...
+    deck_id VARCHAR (255) NOT NULL,
+
+    PRIMARY KEY (tag_id, deck_id),
+    FOREIGN KEY (tag_id) REFERENCES Tag(id),
+    FOREIGN KEY (deck_id) REFERENCES Deck(uuid)
+);
+
+
+CREATE TABLE IF NOT EXISTS User
```
mattyu007: Since we're using Facebook authentication, we'll need to also store [the `userID` returned by the API](https://developers.facebook.com/docs/reference/ios/current/class/FBSDKAccessToken/#userID) so that we can identify our users. In any case, I think it's still a good idea to keep an additional UUID for each user should we decide to switch to a different authentication mechanism in the future.

mattyu007 (commented):
Great first pass. We still need to add support for storing flags.

Server/SQL/schema.sql
ClaytonPassmore: Nope! I'll change that

Server/SQL/schema.sql
ClaytonPassmore: sounds good!

Server/SQL/schema.sql
ClaytonPassmore: Sure, I'm cool with that!

Server/SQL/schema.sql
ClaytonPassmore: ok, let's make it `NOT NULL`

Server/SQL/schema.sql
ClaytonPassmore: lol good catch

Server/SQL/schema.sql
ClaytonPassmore: I think we could have a fourth option where public is false and share_code is a string..

Server/SQL/schema.sql
ClaytonPassmore: 👍

Server/SQL/schema.sql
ClaytonPassmore: yes

Server/SQL/schema.sql
ClaytonPassmore: Ok that's fine

Server/SQL/schema.sql
ClaytonPassmore: I don't think we actually will ever be deleting the actual deck ID, will we? But yeah, I can add it just in case.

henryc132 (approved):

PR #4, opened 2017-02-15
Schema Second Pass

ClaytonPassmore:
- Added datetime fields for creation and most recent update
- Added versioning for users in the library table to track user-specific deck metadata
- Added column for last device to modify a deck in a user's library

### DIFF
--- a/Server/SQL/schema.sql
+++ b/Server/SQL/schema.sql
@@ -1,18 +1,24 @@
+CREATE DATABASE Cue;
+USE Cue;
+
+
 CREATE TABLE IF NOT EXISTS User
 (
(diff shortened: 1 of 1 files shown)


ClaytonPassmore:
I couldn't use `DEFAULT NOW()` for two different columns in the same table. Apparently that's a MySQL limitation. It should be fine though.

Let me know if we should maybe use a skinnier column for the `lastest_modifier`

ClaytonPassmore:
I also expanded the size of the access token because Facebook's access tokens are ~160 characters, so I went to 200 just to be safe. I'm open to changing it.

PR #11, opened 2017-02-19
Add delete flag to metadata repsonse

ClaytonPassmore:
@mattyu007 is this okay?

### DIFF
--- a/Spec/API.md
+++ b/Spec/API.md
@@ -188,7 +188,7 @@ Request description:
 Response:
 ```
 {
-    user-data-version: <user-data-version>
+    user_data_version: <user-data-version>
 }
 ```
 Here, `<user-data-version>` indicates the new version of the user data for the deck.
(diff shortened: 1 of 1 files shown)


Spec/API.md
```diff
@@ -355,7 +355,8 @@ This response only contains metadata (does not include the cards in the deck).
     "created": "<created>"
```
mattyu007, line 355: GitHub won't let me select this, but should `user_data-version` be `user_data_version`?

Spec/API.md
```diff
@@ -373,6 +374,7 @@ Response description:
 - `last_update`: ISO format UTC datetime string describing the time of the most recent deck update
 - `last_update_device`: the name of the device that last modified the deck (this is only available for decks that are owned by the requester)
 - `share_code`: the code that can be entered by users to gain access to a remote private deck
+- `deleted`: boolean indicating whether or not a deck has been deleted by the owner
```
mattyu007, line 377: If `deleted` is true, should the client still expect to receive all the other metadata?

Spec/API.md
mattyu007: Ditto with `PUT /api/v1/deck/<uuid>/flag`, where
```
 {
     user-data-version: <user-data-version>
 }
 ```
should be
```
 {
     "user_data_version": <user-data-version>
 }
 ```

Feel free to address these in a separate PR.

mattyu007 (approved):

Spec/API.md
ClaytonPassmore: Probably not. I'll make a note

Spec/API.md
ClaytonPassmore: Thanks for catching that, I'll throw those in here too

PR #12, opened 2017-02-20
API endpoints for discover

henryc132:


Server/App/api/discover.py
```diff
@@ -1,35 +1,66 @@
...
-
 class TopHandler(BaseHandler):
     @authorize_request_and_create_db_connector
-    def get(self, connector):
+    @require_query_string_params('page')
+    def get(self, page, connector):
         """Retrieve metadata for the most popular decks."""
-        self.abort(404)
+
+        """Retrieve a page of deck's metadata"""
```
ClaytonPassmore: Nit: can you change this from a doc string to a regular comment?

Server/App/api/discover.py
```diff
@@ -1,35 +1,66 @@
...
+
+        # Return search results to the user.
+        return self.make_response(response=json.dumps(decks, default=serializer),
+            status=StatusCode.OK)
 
 class SearchHandler(BaseHandler):
     @authorize_request_and_create_db_connector
-    @require_query_string_params('query')
-    def get(self, query, connector):
+    @require_query_string_params('query', 'tags', 'page')
```
ClaytonPassmore, line 45: Would you mind adding any parameters that you added to the API doc?

Server/App/api/discover.py
```diff
@@ -1,35 +1,66 @@
...
+    def get(self, query, tags_json, page, connector):
         """Retrieve metadata for decks matching the search string."""
-        self.abort(404)
 
+	tags = json.loads(tags_json)
+
+        tags_string = ",".join(tags)
```
ClaytonPassmore: I think this line is indented one space too far

Server/SQL/schema.sql
```diff
@@ -15,7 +15,7 @@ CREATE TABLE IF NOT EXISTS Deck
 (
     uuid VARCHAR(36) NOT NULL PRIMARY KEY,
     created TIMESTAMP NOT NULL DEFAULT NOW(),
-    last_update TIMESTAMP NOT NULL,
+    last_update TIMESTAMP NULL,
```
ClaytonPassmore, line 18: Why did you change this?

henryc132: Did this line work for you when you built? Its was set as not null, but given no default, so it wouldn't build for me. And since we can't set it to Now() because of limitations, I figured we might as well let it be null

Server/SQL/procedures.sql
```diff
@@ -263,5 +307,34 @@ BEGIN
...
+        SELECT version, last_update_device, deck_id
+        FROM Library
+    ) AS L ON L.deck_id=Deck.uuid
+    WHERE Deck.public=TRUE
+    AND (Deck.name like CONCAT('%',query_string,'%')
+        OR Deck.uuid IN (SELECT dt.deck_id FROM DeckTag dt
+	    LEFT JOIN Tag t ON dt.tag_id=t.id
+	    WHERE FIND_IN_SET(t.tag,tags)
+	    GROUP BY dt.deck_id
+	    HAVING COUNT(dt.tag_id) = num_tags));
```
ClaytonPassmore: Should we maybe order by rating as well?

Server/SQL/procedures.sql
```diff
@@ -263,5 +307,34 @@ BEGIN
...
+        Deck.created,
+        Deck.last_update,
+        L.last_update_device,
+        Deck.share_code
+    FROM Deck LEFT JOIN (
+        SELECT version, last_update_device, deck_id
+        FROM Library
+    ) AS L ON L.deck_id=Deck.uuid
+    WHERE Deck.public=TRUE
+    AND (Deck.name like CONCAT('%',query_string,'%')
```
ClaytonPassmore: I think for all string matches we should make it case insensitive

Server/SQL/schema.sql
ClaytonPassmore: If you use the `CREATE_DECK` procedure, it will fill this in with `NOW()`

Server/App/api/discover.py
```diff
@@ -1,35 +1,66 @@
...
 class TopHandler(BaseHandler):
     @authorize_request_and_create_db_connector
-    def get(self, connector):
-        """Retrieve metadata for the most popular decks."""
-        self.abort(404)
+    @require_query_string_params('page')
+    def get(self, page, connector):
+        #Retrieve metadata for the most popular decks.
+
+        #Retrieve a page of deck's metadata"""
```
ClaytonPassmore: Trailing `"""`

Server/App/api/discover.py
```diff
@@ -1,35 +1,66 @@
...
 
-
 class TopHandler(BaseHandler):
     @authorize_request_and_create_db_connector
-    def get(self, connector):
-        """Retrieve metadata for the most popular decks."""
-        self.abort(404)
+    @require_query_string_params('page')
+    def get(self, page, connector):
+        #Retrieve metadata for the most popular decks.
```
ClaytonPassmore: This can stay as a doc string

Server/App/api/discover.py
ClaytonPassmore: Also add a space after the `#`

Server/SQL/procedures.sql
```diff
@@ -263,5 +307,35 @@ BEGIN
...
+        Deck.share_code
+    FROM Deck LEFT JOIN (
+        SELECT version, last_update_device, deck_id
+        FROM Library
+    ) AS L ON L.deck_id=Deck.uuid
+    WHERE Deck.public=TRUE
+    AND (LOWER(Deck.name) like CONCAT('%',LOWER(query_string),'%')
+        OR Deck.uuid IN (SELECT dt.deck_id FROM DeckTag dt
+	    LEFT JOIN Tag t ON dt.tag_id=t.id
+	    WHERE FIND_IN_SET(t.tag,tags)
```
ClaytonPassmore: I think we need to do lowercase on tags too... Or should we enforce that all tags must always be lowercase so that we don't have to do it here?

henryc132: I think we enforce that we save it in lower?

henryc132: But theres no extra cost in using lower here anyways

ClaytonPassmore: ok!

ClaytonPassmore: Well find_in_set isn't case insensitive right? So we would have to make things lowercase first... So instead of adding that we can just force all tags to be lowercase like you said.

Server/App/api/discover.py
```diff
@@ -1,35 +1,66 @@
...
+    @require_query_string_params('query', 'tags', 'page')
+    def get(self, query, tags_json, page, connector):
         """Retrieve metadata for decks matching the search string."""
-        self.abort(404)
 
+        tags = json.loads(tags_json)
+
+        tags_string = ",".join(tags)
+
+	decks = connector.call_procedure('SEARCH_DECKS', query, tags_string, len(tags), page)
```
ClaytonPassmore: There's some weird indentation going on here.. lines 53, 55, and 59 need to be indented by one more space.

Server/SQL/procedures.sql
```diff
@@ -263,5 +307,35 @@ BEGIN
...
+        Deck.share_code
+    FROM Deck LEFT JOIN (
+        SELECT version, last_update_device, deck_id
+        FROM Library
+    ) AS L ON L.deck_id=Deck.uuid
+    WHERE Deck.public=TRUE
+    AND (LOWER(Deck.name) like CONCAT('%',LOWER(query_string),'%')
+        OR Deck.uuid IN (SELECT dt.deck_id FROM DeckTag dt
+	    LEFT JOIN Tag t ON dt.tag_id=t.id
+	    WHERE FIND_IN_SET(LOWER(t.tag),LOWER(tags))
```
ClaytonPassmore, line 334: Can you apply `lower()` to a set of tags?

ClaytonPassmore:
@henryc132 just a couple of minor comments.

Server/App/api/discover.py
henryc132: Well that is annoying, look perfectly fine on my editor. I'll try and see whats happening

Server/SQL/procedures.sql
henryc132: Tags is simply a comma delimited string, so LOWER is fine in this case

ClaytonPassmore: true (y)

Server/App/api/discover.py
ClaytonPassmore: could be tabs vs spaces?

henryc132:
@ClaytonPassmore Alright, fixed those darn spacing issues

ClaytonPassmore (approved):

ClaytonPassmore:
@henryc132 LGTM, merge away

PR #16, opened 2017-02-20
Adding v1 UX mocks PDF

mattyu007:
Finally, v1 mocks for the overall Cue UX. This PDF includes most screens for both Android and iOS, though is currently missing (most notably):

* Sync Conflict Resolution flow
* Transfer Ownership flow

Having gone through this UX exercise, I think we'll have a better feel for how much work still needs to be done on the app. On some screens, the visual design differs quite a bit to better match the platform's design language, but we should still be able to share a large portion of our UI code between Android and iOS.

On certain pages, you'll see **TBD** in red all-caps; these are places where I still have outstanding questions which need to be addressed before we'll be able to finish our app.


ClaytonPassmore:
@mattyu007 in terms of the ownership transfer, I'm not sure how we're going to do that since users don't really have any concept of what other users' user IDs are.

ClaytonPassmore (approved):

ClaytonPassmore:
@mattyu007 GTM

mattyu007:
@ClaytonPassmore Re: User IDs: We're going to run into that exact issue when we display public decks or shared decks in either the Library or in the Discovery flows since the metadata response only returns a user ID, and the app has no way to turn that into a displayable name.

PR #17, opened 2017-02-20
Added number of bytes for token to config

ClaytonPassmore:
Forgot to add this on initial server commit.

### DIFF
--- a/Server/App/config.py
+++ b/Server/App/config.py
@@ -45,3 +45,4 @@ class FacebookConfig(object):
 class SecurityConfig(object):
     ACCESS_TOKEN_HEADER = 'X-CUE-ACCESS-TOKEN'
     USER_ID_HEADER = 'X-CUE-USER-ID'
+    ACCESS_TOKEN_RAW_BYTES = 120


ClaytonPassmore:
@henryc132 can you take a look?

Server/App/config.py
```diff
@@ -45,3 +45,4 @@ class FacebookConfig(object):
 class SecurityConfig(object):
     ACCESS_TOKEN_HEADER = 'X-CUE-ACCESS-TOKEN'
     USER_ID_HEADER = 'X-CUE-USER-ID'
+    ACCESS_TOKEN_RAW_BYTES = 120
```
henryc132, line 48: Is FB access token only 120 bytes, thought it was larger

henryc132: Actually I guess you mean the access token we generate to give back to user, but that is a 160 character VARCHAR in schema?

henryc132 (approved):

Server/App/config.py
ClaytonPassmore: Yeah so it generates 120 bytes but when you encode it in base 64 it ends up being 160

PR #18, opened 2017-02-21
Add additional paramater for discoverability endpoint

henryc132:


Spec/API.md
```diff
@@ -355,9 +355,12 @@ Response:
 ### Search
 Request:
 ```
-GET /api/v1/discover/search?query=<query>[&page=<page>]
+GET /api/v1/discover/search?query=<query>&tags=<tags>[&page=<page>]
 ```
```
henryc132, line 359: Separated the title search from the tag search

Spec/API.md
```diff
@@ -339,7 +339,7 @@ Request:
...
 GET /api/v1/discover/new[?page=<page>]
 ```
-Where `page` is the page offset into the result set (0 by default).
+Where `page` is the page offset into the result set (1 by default).
```
henryc132, line 343: Changed to match code, default page is 1. Thought it made more sense to start at 1 in this case.

ClaytonPassmore:
@henryc132 is the `page` parameter optional or required? If it's required, you should drop the square brackets around it in the request. Other than that, looks good

ClaytonPassmore (approved):

henryc132:
Its optional, you just get return first page results if you don't specify it

PR #19, opened 2017-02-21
Initial react-native app

Tyler-Brenneman:
First attempt at writing a react-native app. Simple navigation menu works on android, unable to check if ios tab menu works.

PR #20, opened 2017-02-21
initial react-native app with basic navigation

Tyler-Brenneman:
Remake pull request #19 with proper branches. Start of login screen and navigation menu for library, discover, and search. Tested on android but not ios.


Cue/app/tabs/CueTabsView.ios.js
```diff
@@ -0,0 +1,72 @@
...
+  onTabSelect(tab: Tab) {
+    if (this.props.tab !== tab) {
+      this.props.onTabSelect(tab);
+    }
+  }
+
+  render() {
+    return (
+      <TabBarIOS>
+        <TabBarItemIOS
```
mattyu007: This needs to be `TabBarIOS.Item`.

Cue/app/tabs/CueTabsView.ios.js
```diff
@@ -0,0 +1,72 @@
...
+          title="Library"
+          selected={this.props.tab === 'library'}
+          onPress={this.onTabSelect.bind(this, 'library')}
```
mattyu007: Missing a `>` at the end of this line.

Cue/app/tabs/CueTabsView.ios.js
```diff
@@ -0,0 +1,72 @@
...
+          <LibraryView
+            navigator={this.props.navigator}
+          />
+        </TabBarItemIOS>
+        <TabBarItemIOS
+          title="Discover"
+          selected={this.props.tab === 'discover'}
+          onPress={this.onTabSelect.bind(this, 'discover')}
```
mattyu007: Same here.

Cue/app/tabs/CueTabsView.ios.js
```diff
@@ -0,0 +1,72 @@
...
+          <DiscoverView
+            navigator={this.props.navigator}
+          />
+        </TabBarItemIOS>
+         <TabBarItemIOS
+          title="Search"
+          selected={this.props.tab === 'search'}
+          onPress={this.onTabSelect.bind(this, 'search')}
```
mattyu007: Same here.

Cue/app/common/CueDrawerLayout.js
```diff
@@ -0,0 +1,67 @@
...
+    (this: any).openDrawer = this.openDrawer.bind(this);
+    (this: any).closeDrawer = this.closeDrawer.bind(this);
+    (this: any).onDrawerOpen = this.onDrawerOpen.bind(this);
+    (this: any).onDrawerClose = this.onDrawerClose.bind(this);
+    (this: any).handleBackButton = this.handleBackButton.bind(this);
+  }
+
+  render() {
+    const {drawerPosition, ...props} = this.props;
+    const {Right, Left} = DrawerLayoutAndroid.positions;
```
mattyu007: Strangely, I'm not able to get the app running on iOS until I comment this line (and line 26) out, otherwise I get the error:
```
undefined is not an object (evaluating '_DrawerLayoutAndroid$.Right')
```

My Google-fu isn't turning anything up, and in the interest of saving time, since I don't think we'll ever need to actually have the menu open from the right side, I suggest that we (temporarily) remove this config option until we can come to a solution?

mattyu007 (commented):
Great start, let me know what you think about the drawer position issue.

Cue/app/common/CueDrawerLayout.js
Tyler-Brenneman: I think we can remove, don't need to have it since the menu will always be the same on the left

mattyu007: I agree, I don't really see a reason to change the menu to appear from the right side unless we add support for a RTL language as well.

mattyu007 (approved):

PR #21, opened 2017-02-23
Add Library endpoints

henryc132:


Server/App/api/library.py
```diff
@@ -29,15 +61,59 @@ class LibraryCopyHandler(BaseHandler):
...
 
 
 class LibraryDeckRemoveHandler(BaseHandler):
     @authorize_request_and_create_db_connector
     @extract_user_id
     def delete(self, user_id, connector, uuid):
         """Remove a deck from a user's library and delete it if they are the owner."""
-        self.abort(404)
+
+        connector.call_procedure('LIBRARY_REMOVE', user_id, uuid)
```
ClaytonPassmore, line 114: If the deck is owned by the user that is deleting it, we also have to mark the deck as deleted and delete all of its cards and tags

ClaytonPassmore:
Overall looks great! Just one case we need to cover where the user deleting the deck also owns the deck.

Server/App/api/library.py
ClaytonPassmore: Also need to remove the needreview entries.

henryc132:
@ClaytonPassmore Updated the library_remove procedure

ClaytonPassmore (approved):
Cool, looks good!

Issue #25, opened 2017-02-24
Resolve rotation issues with height of `CueHeader` and the native toolbars


mattyu007:
On Android and iOS, the height of the status bar + main navigation bar depends on the device orientation, which our custom `CueHeader` does not currently account for.

| Orientation | iOS  | Android |
|-------------|------|---------|
| Portrait    | 20pt + 44pt | 24pt + 56dp    |
| Landscape   | **0pt** + 44pt or 32pt* | 24pt + **48dp**    |

\* Only on ≤ 4" devices (iPhone 5s, 5, 4s). I don't think it's necessary to get this 100%, but iOS hides the status bar in landscape which we need to account for so the header doesn't take up any extra space.

## Screenshots - Android
![android_portrait](https://cloud.githubusercontent.com/assets/13400887/23301648/fb75d4a6-fa58-11e6-82c9-0cd413358768.png)
![android_landscape](https://cloud.githubusercontent.com/assets/13400887/23301650/fec40d08-fa58-11e6-888d-082c2f9f8bf1.png)

## Screenshots - iOS
![ios_portrait](https://cloud.githubusercontent.com/assets/13400887/23301653/0420e4ec-fa59-11e6-96d6-1ce728630fc9.png)
![ios_landscape](https://cloud.githubusercontent.com/assets/13400887/23301654/0431f246-fa59-11e6-9e2b-fdaefbdad3a6.png)

PR #26, opened 2017-02-24
Initial in-tab navigation and look & feel setup

mattyu007:
## In summary, this pull request:

1. begins scaffolding out the basic in-tab navigation structure and app look and feel
2. renames the `navigation` reducer and actions to `tabs` and excludes them from persistence

## In a little more detail:
I probably wasted about 1 1/2 days dabbling with `NavigationExperimental` to try to get that to work before reverting back to the standard `Navigator`. (`Navigator` is EOL now and will be replaced soon by `NavigationExperimental`.)

With this new navigation structure, once you drill down in a tab, you are no longer able to switch tasks until you back all the way out to the top of the tab. That is, on iOS, the tab bar disappears, and on Android, there is no way to reach the drawer menu. Additionally, each scene is responsible for drawing its own navigation bar / toolbar. This gives us much more control especially if we want to do anything fancy with the navigation bar later on. wink wink.

I also pulled in some Material Design icon assets for us to use.

Outstanding nav / L&F tasks or investigation items:
* `CueHeader` still needs to be augmented with the ability to have custom objects drawn inside it. Mainly this is to support having the search box appear inside the header, but if this becomes too much to handle, we can just adjust the design and have the search box appear under the header.
* I'm still not sure if doing platform detection and adjusting the header is the right way to go (see `LibraryHome.render()`). However, in most cases, there isn't enough of a difference to justify actually splitting the component into `.android.js` and `.ios.js` counterparts.
* There's some funky funky going on with rotation and the height of the header which I need to look into - #25


Tyler-Brenneman:
Looks good!

I think the platform detection in LibraryHome.render() is fine for small differences like this. the rest of the component should be platform independent so we should be good.

Tyler-Brenneman (approved):

PR #29, opened 2017-02-25
CueHeader iOS refinements

mattyu007:
- Increase minimum touch target size to 44x44pt
- Prevent long text from wrapping

home-before
home-after


Tyler-Brenneman (approved):
looks good!

Issue #30, opened 2017-02-25
Error on authentication api


Tyler-Brenneman:
Error when returning existing credentials in authentication api
```
Traceback (most recent call last):
[2017-02-25 16:25:21,639] ERROR in app: Exception on /api/v1/auth [POST]
  File "/home/tyler/gits/ProjectOrange/Server/venv/local/lib/python2.7/site-packages/flask/app.py", line 1982, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/tyler/gits/ProjectOrange/Server/venv/local/lib/python2.7/site-packages/flask/app.py", line 1614, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/tyler/gits/ProjectOrange/Server/venv/local/lib/python2.7/site-packages/flask/app.py", line 1517, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/home/tyler/gits/ProjectOrange/Server/venv/local/lib/python2.7/site-packages/flask/app.py", line 1612, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/tyler/gits/ProjectOrange/Server/venv/local/lib/python2.7/site-packages/flask/app.py", line 1598, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/tyler/gits/ProjectOrange/Server/App/utils/wrappers.py", line 88, in wrapper
    return func(self, *(param_vals + list(args)[1:]), **kwargs)
  File "/home/tyler/gits/ProjectOrange/Server/App/api/authentication.py", line 37, in post
    'user-id': result['uuid'],
TypeError: list indices must be integers, not str
127.0.0.1 - - [25/Feb/2017 16:25:21] "POST /api/v1/auth HTTP/1.1" 500 -
INFO:werkzeug:127.0.0.1 - - [25/Feb/2017 16:25:21] "POST /api/v1/auth HTTP/1.1" 500 -
```

PR #31, opened 2017-02-25
Bug/authentication api

Tyler-Brenneman:
fix for issue #30. Was getting the following error when re authenticating
```
File "/home/tyler/gits/ProjectOrange/Server/App/api/authentication.py", line 37, in post
    'user-id': result['uuid'],
TypeError: list indices must be integers, not str
```


henryc132 (approved):
Looks good

PR #32, opened 2017-02-26
implement facebook login and authentication with cue server

Tyler-Brenneman:
Authenticates with facebook and cue server. 

When authenticated successfully with cue server, the userId and accessToken will be stored in the user store.

Once the login is complete, an asynchronous request will be sent to facebook graph api to update the name in the user store.

Created a folder Api where we can put all of our different api calls, this way it will follow a similar structure as the server and separate the calls.

An outstanding item is when a 201 response is received from the cue server from authenticating, we should send a request to send the user's name on the server. On a 200 response we could query the cue server for the username instead of facebook. This detail doesn't matter that much, as it works fine now with just getting the user name from facebook's graph api.

Also I only did the facebook-sdk setup for android. The setup for ios requires xcode and I dont have.
Could Somebody with xcode following the ios setup instructions here: 
https://github.com/facebook/react-native-fbsdk


Cue/app/api/Authentication.js
```diff
@@ -0,0 +1,35 @@
...
+		.then(checkStatus)
+		.then(response => response.json())
+		.catch(e => {
+			throw (e)
+		});
+	},
+}
+
+var checkStatus =function(response) {
+	if (response.status >= 200 || response.status <= 300) {
```
mattyu007: This should be `&& response.status < 300` instead of `|| response.status <= 300`

mattyu007 (approved):

Cue/app/api/Authentication.js
```diff
@@ -0,0 +1,35 @@
...
+		.then(checkStatus)
+		.then(response => response.json())
+		.catch(e => {
+			throw (e)
+		});
+	},
+}
+
+var checkStatus = function(response) {
+	if (response.status >= 200 || response.status < 300) {
```
mattyu007: This needs to be `&&`

mattyu007 (changes requested):
Wait, just one tiny thing

mattyu007 (approved):

PR #33, opened 2017-02-27
Initial version of library list view

mattyu007:
Initial version of the library list view, including empty and filled state.

Note: This version does not support displaying the corner insets for shared and public decks yet.

empty
filled


Cue/app/tabs/library/LibraryListView.js
```diff
@@ -3,30 +3,122 @@
...
+  public: boolean,
+  deckVersion: number,
+  userDataVersion: number,
+  created: Date,
+  lastUpdate: Date,
+  lastUpdateDevice?: string,
+  shareCode?: string,
+  deleted: boolean,
+  cards: Array<Card>
+}
```
Tyler-Brenneman: Agreed these types should be moved to a common class. api/types.js would work. This way the api can easily use and return these types

mattyu007: Good call. Let's get #38 in first then and update this PR before merging this.

mattyu007:
@lonebutterfish I updated the member names for the types in `app/api/types.js` to reflect what comes out from the API

Tyler-Brenneman (approved):

PR #36, opened 2017-03-01
Initial version of Discover tab with placeholder views

mattyu007:
Extremely rudimentary and preliminary version of the basic layout of the Discover tab with placeholders where decks should be. Opening this PR now so the Discover tab isn't completely empty during the demo.

discover_placeholders


Tyler-Brenneman (dismissed):

Tyler-Brenneman (approved):
Good start.

PR #37, opened 2017-03-01
Add tab icons

mattyu007:
Hooray, tab icons! 😀 Icons and selected states are styled (mostly) according to platform conventions, of course.

iOS is still missing an "Account" icon, but that's not necessary for the demo, so it's not included here yet.

## Screenshots
1-library
2-discover
3-search


Tyler-Brenneman (approved):
looks good!

PR #38, opened 2017-03-01
Feature/fetch library

Tyler-Brenneman:
Implemented library synchronization on login.
After cue auth has finished, the library will be retrieved asynchronously. 

state.library.decks is null if synchronization hasn't finished yet (not sure if this is how we want to handle this)
state.library.decks size is 0 if library is empty

The code is a little messy but should work for the demo. I was unsure of the best practice on getting state from the api and how to use the types.

### DIFF
--- a/Cue/app/actions/library.js
+++ b/Cue/app/actions/library.js
@@ -1,0 +1,13 @@
+// @flow
+import type { PromiseAction } from './types';
+import LibraryApi from '../api/Library';
+
+async function loadLibrary(state): PromiseAction {
+  let library = await LibraryApi.fetchLibrary(state);
+  return {
+    type: 'LOADED_LIBRARY',
+    library,
+  };
+}
+
+module.exports = { loadLibrary };
--- a/Cue/app/api/Library.js
+++ b/Cue/app/api/Library.js
@@ -1,0 +1,36 @@
+// @flow
+import { serverURL } from '../env';
+
+
+var LibraryApi = {
+	//TODO: not sure proper way to get state in the api
+	//TODO: look into moving checkStatus and auth header creation into common util
+	//TODO: return Array<Deck> instead of text
+	fetchLibrary(state) {
+		let url = serverURL + '/api/v1/library';
+		let user = state.user;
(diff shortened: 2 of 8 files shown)


Cue/app/actions/library.js
```diff
@@ -0,0 +1,13 @@
+// @flow
+import type { PromiseAction } from './types';
+import LibraryApi from '../api/Library';
+
+async function loadLibrary(state): PromiseAction {
+  let library = eval(await LibraryApi.fetchLibrary(state));
```
mattyu007: Why do we need this `eval`? Could we not just do `response => response.json()` in line 19 of `LibraryApi`?

I'm concerned that using `eval` on the result of a network call could be a security risk.

Cue/app/api/Library.js
```diff
@@ -0,0 +1,36 @@
+// @flow
+import { serverURL } from '../env';
+
+
+var LibraryApi = {
+	//TODO: not sure proper way to get state in the api
```
mattyu007, line 6: When we add more API calls in the future, we could look at adding some static properties to, say, `AuthenticationApi` to avoid needing to pass the entire app state into every API call.

mattyu007 (changes requested):

Cue/app/api/Library.js
Tyler-Brenneman: That sounds like a good plan. Ill make an issue for this and assign to myself, but don't think we need this change for the demo.

Cue/app/actions/library.js
Tyler-Brenneman: for some reason when i was testing it last night and response.json() was returning undefined. It was late and couldnt figure out why but text() was working.

Trying it now and response.json() works so ill make the change to use that.

mattyu007: OK. If you were testing on `projectorange.club` it might have been because the DB stored procedures weren't updated, so the server was returning bad results, according to @ClaytonPassmore.

mattyu007 (approved):

PR #41, opened 2017-03-02
Fetch deck with cards

Tyler-Brenneman:
adds the 'GET /api/v1/deck/<uuid>' request

On login it will now fetch all decks and cards and stores it in store.library.decks

### DIFF
--- a/Cue/app/api/Library.js
+++ b/Cue/app/api/Library.js
@@ -2,12 +2,28 @@
 import { serverURL } from '../env';
 
 
+//TODO: not sure proper way to get state in the api
+//TODO: look into moving checkStatus and auth header creation into common util
 var LibraryApi = {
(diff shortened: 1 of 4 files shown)


Cue/app/api/Library.js
```diff
@@ -21,6 +21,22 @@ var LibraryApi = {
 			throw (e)
 		});
 	},
+
+	fetchDeck(uuid : string, state) {
+		let url = serverURL + '/api/v1/deck/' + uuid;
+		let user = state.user;
+		return fetch(url, {
```
mattyu007, line 28: When we have some time, we should consider refactoring the existing API classes to use a common base method which just takes in the API route and method, handles authentication and error checking, and returns a `Promise`. I feel like that would save us a lot of redundant code.

mattyu007 (approved):

PR #43, opened 2017-03-02
Initial version of deck view

mattyu007:
Pretty rough first pass of the deck view screen. Most of the styling is still modelled after iOS, so many bits look out of place on Android. Works though. ┐('～`；)┌ 

## Working Features
* deck title display
* card display
* header insets and subtitles for shared and public decks, except that the other owner's name can't be displayed

## Screenshots
Basic Private Deck
private

Shared Deck
shared

Public Deck
public


Tyler-Brenneman (dismissed):

Cue/app/tabs/library/DeckView.js
```diff
@@ -41,18 +57,69 @@ export default class DeckView extends React.Component {
...
+
+    let overflowItems
+    if (Platform.OS === 'android') {
+      overflowItems = [
+        {
+          title: 'Show flagged cards only'
+        }
+      ]
+    }
+
```
Tyler-Brenneman, line 96: This can be moved up into above if check, but doesn't really matter if we are going to split into 2 files(Which I agree we should do)

Cue/app/tabs/library/DeckView.js
```diff
@@ -12,26 +12,42 @@ import type { Card, Deck } from '../../api/types'
...
 }
 
-export default class DeckView extends React.Component {
-  props: {
-    navigator: Navigator,
-    deck: Deck
+type Props = {
+  navigator: Navigator,
+  deck: Deck,
+  userId: string
```
Tyler-Brenneman, line 31: curious why userId is passed in as a prop. Does the deckview need to know userId to tell if the deck is owned by them?

Tyler-Brenneman: I see it is used by DeckViewInfoHeader.js, but only the deck is being passed in as a prop right now on line 104, so userId wont be set in DeckViewInfoHeader.

Tyler-Brenneman: nevermind ignore these comments, i see how this works now. The connect from react-redux sets the userId prop.

Tyler-Brenneman (approved):
Looks good! I got a little confused in how react-redux sets the state in the deckview, but it looks good to me now!

Cue/app/tabs/library/DeckView.js
mattyu007: Yeah unfortunately the `userId` is required to know whether we show the deck insets / sharing status subtitles. They're filled in using Redux, I'll add comments in the next PR.

Cue/app/tabs/library/DeckView.js
mattyu007: For sure, this is definitely going to be split into `.ios` and `.android` components.

PR #44, opened 2017-03-02
Layout and graphics on login screen

mattyu007:
Since we're going to be demoing the login screen as well, I gave it a quick coat of paint. The layout and icon still look kinda wonky, but we can tidy that up later.

login


Tyler-Brenneman (dismissed):

Tyler-Brenneman (approved):
Looks good. Only thing i noticed is the "Login with Facebook" button looks a bit cutoff on the ios pic. Not sure if its still like that since I cant test it. Very minor detail and not sure if we want to look into.

mattyu007:
Yeah, the button is cut off on iOS for some reason. I don't think it's a very high priority item for us to look at right now though.

Issue #14, opened 2017-02-20
Add handler for deck overwrites


ClaytonPassmore:
See API spec for details.

Issue #9, opened 2017-02-19
Add share code generator handler


ClaytonPassmore:
Add a handler that generates deck share codes in `api/deck.py`.

PR #49, opened 2017-03-05
Fix issues #8,#9,#10,#14,#47 

### LINKED ISSUE #8: Set last_update and last_update_device when modifying decks
This needs to happen in `api/deck.py` whenever we increment the deck version or the user data version.

henryc132:
Fixes #8, #9, #10, #14, #47 

Fix Includes:

- Set last_update and last_update_device 
- Add share code generator handle
- Store tags in lowercase
- Add owner name to metadata

- Add overwrite endpoint

### DIFF
--- a/Server/App/api/deck.py
+++ b/Server/App/api/deck.py
@@ -5,6 +5,8 @@ Also, handle the flagging of cards for review.
 
 import json
 import logging
+import string
+import random
 
 from config import StatusCode
 from utils.base_handler import BaseHandler
@@ -84,9 +86,9 @@ class DeckUUIDHandler(BaseHandler):
 
     @authorize_request_and_create_db_connector
     @extract_user_id
-    @require_params('parent_deck_version', 'parent_user_data_version')
+    @require_params('parent_deck_version', 'parent_user_data_version', 'device')
     @optional_params('name', 'public', 'tags', 'actions')
-    def put(self, name, public, tags, actions, parent_deck_version, parent_user_data_version,
+    def put(self, name, public, tags, actions, parent_deck_version, parent_user_data_version, device,
             user_id, connector, uuid):
         """Edit an existing deck."""
 
@@ -168,7 +170,7 @@ class DeckUUIDHandler(BaseHandler):
             # Increment both the deck version and the user data version.
             # TODO set last_update and last_update_device.
             connector.call_procedure('INCREMENT_DECK_VERSION', user_id, uuid)
-            connector.call_procedure('INCREMENT_USER_DATA_VERSION', user_id, uuid)
+            connector.call_procedure('INCREMENT_USER_DATA_VERSION', user_id, uuid, device)
 
             # Commit changes.
             connector.end_transaction()
@@ -194,8 +196,8 @@ class DeckUUIDHandler(BaseHandler):
 class DeckFlagHandler(BaseHandler):
     @authorize_request_and_create_db_connector
     @extract_user_id
-    @require_params('parent_deck_version', 'parent_user_data_version', 'actions')
-    def put(self, parent_deck_version, parent_user_data_version, actions, user_id, connector, uuid):
+    @require_params('parent_deck_version', 'parent_user_data_version', 'actions', 'device')
+    def put(self, parent_deck_version, parent_user_data_version, actions, device, user_id, connector, uuid):
         """Flag specific cards in an existing deck."""
 
         versions = connector.call_procedure('GET_VERSIONS', user_id, uuid)
@@ -221,7 +223,7 @@ class DeckFlagHandler(BaseHandler):
 
             # Increment the user data version and retrieve it.
             new_user_data_version = connector.call_procedure(
-                'INCREMENT_USER_DATA_VERSION', user_id, uuid)[0]
+                'INCREMENT_USER_DATA_VERSION', user_id, uuid, device)[0]
 
             # Commit the changes.
             connector.end_transaction()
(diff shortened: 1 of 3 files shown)


Server/App/api/deck.py
```diff
@@ -168,7 +170,7 @@ def put(self, name, public, tags, actions, parent_deck_version, parent_user_data
             # Increment both the deck version and the user data version.
             # TODO set last_update and last_update_device.
             connector.call_procedure('INCREMENT_DECK_VERSION', user_id, uuid)
-            connector.call_procedure('INCREMENT_USER_DATA_VERSION', user_id, uuid)
+            connector.call_procedure('INCREMENT_USER_DATA_VERSION', user_id, uuid, device)
```
ClaytonPassmore, line 173: I'm wondering if we should also add the device to the create deck procedure as well

Server/SQL/procedures.sql
```diff
@@ -149,21 +155,34 @@ END$$
...
 -- Increment deck version number. It will only be incremented if the user owns the deck.
 CREATE PROCEDURE INCREMENT_DECK_VERSION(IN uid VARCHAR(36), IN did VARCHAR(36))
 BEGIN
-    UPDATE Deck SET version=(version + 1) WHERE owner=uid AND uuid=did;
+    UPDATE Deck SET version=(version + 1), last_update=NOW() WHERE owner=uid AND uuid=did;
     SELECT version FROM Deck WHERE owner=uid AND uuid=did;
 END$$
 
 
+-- Generate share code for deck.
```
ClaytonPassmore: Maybe say 'Insert a new share code for a deck if it doesn't already have one'

Server/App/api/deck.py
```diff
@@ -190,12 +192,105 @@ def put(self, name, public, tags, actions, parent_deck_version, parent_user_data
...
+        if len(versions) != 1:
+            return self.make_response(status=StatusCode.NOT_FOUND)
+
+        # Begin deck edits
+        try:
+            connector.begin_transaction()
+
+            # Collect and validate metadata updates.
+            metadata_edits = {}
+            if name is not None:
```
ClaytonPassmore: Since this is required, you can remove this line and the `isinstance` check should catch a none type

Server/App/api/deck.py
```diff
@@ -190,12 +192,105 @@ def put(self, name, public, tags, actions, parent_deck_version, parent_user_data
...
+                if not isinstance(name, (str, unicode)):
+                    raise TypeError
+                metadata_edits['name'] = name
+            if public is not None:
```
ClaytonPassmore: same here

Server/App/api/deck.py
```diff
@@ -190,12 +192,105 @@ def put(self, name, public, tags, actions, parent_deck_version, parent_user_data
...
+                    raise TypeError
+                metadata_edits['public'] = public
+
+            # Update metadata.
+            query_str = 'UPDATE Deck SET {key}=%s WHERE uuid=%s'
+            for k, v in metadata_edits.iteritems():
+                connector.query(query_str.format(key=k), v, uuid)
+
+            # Update tags.
+            if tags is not None:
```
ClaytonPassmore: Instead, maybe do
```Python
if not isinstance(tags, list):
    raise TypeError()
```

Server/App/api/deck.py
```diff
@@ -256,10 +351,30 @@ def post(self, rating, user_id, connector, uuid):
...
+    @authorize_request_and_create_db_connector
+    @extract_user_id
+    def get(self, user_id, connector, uuid):
+        results = connector.call_procedure('FETCH_DECK', user_id, uuid)
+        if len(results) != 1:
+            return self.make_response(status=StatusCode.INTERNAL_SERVER_ERROR)
+        if results[0]['owner'] != user_id:
+            return self.make_response(status=StatusCode.FORBIDDEN)
+
+        while True:
```
ClaytonPassmore, line 365: Have you tested this with a collision? You _might_ need a `try`/`except` around the `call_procedure`... I think it might throw an exception if it isn't unique.

ClaytonPassmore (changes requested):
Would you be able to document the new API endpoints in the API doc?

Server/App/api/deck.py
henryc132: The stored procedure won't actually try to insert if it isn't unique

ClaytonPassmore: Ok that's true, I missed that. However, I think there will still be an issue because when there is a collision, it will return a single result of `null`, so `len(set_code) == 1`, even if a new code was not set.

henryc132:
@ClaytonPassmore made changes, also the overwrite and share code endpoints are both already in the api doc

Server/App/api/deck.py
```diff
@@ -190,12 +192,106 @@ def put(self, name, public, tags, actions, parent_deck_version, parent_user_data
...
+        if len(versions) != 1:
+            return self.make_response(status=StatusCode.NOT_FOUND)
+
+        # Begin deck edits
+        try:
+            connector.begin_transaction()
+
+            # Collect and validate metadata updates.
+            metadata_edits = {}
+            if name is None:
```
ClaytonPassmore: Maybe do
```Python
if not isinstance(name (str, unicode)):
    ...
```

Server/App/api/deck.py
```diff
@@ -190,12 +192,106 @@ def put(self, name, public, tags, actions, parent_deck_version, parent_user_data
...
+                raise TypeError
+                
+            if public is None:
```
ClaytonPassmore: and here do
```Python
if not isinstance(public, bool):
    ...
```

ClaytonPassmore:
@henryc132 yup sorry, I'm blind :laughing:

henryc132:
Yea makes sense, added those type checks

ClaytonPassmore:
@henryc132 looks good! You don't really need the none check in `if name is None or not isinstance...` because the isinstance will be false if the variable is of none type. But it doesn't really matter :stuck_out_tongue:

ClaytonPassmore (approved):

ClaytonPassmore:
:boat:

PR #50, opened 2017-03-06
Added Make target so stored procedures can be reloaded

ClaytonPassmore:
- New target for reloading procedures (now we don't have to drop the DB each time).
- Miscellaneous code cleanup
- Require DB password on schema update and procedure update
- Allow DB username, password, and host to be overwritten with environment variables (but keep the same defaults for developing)

### DIFF
--- a/Server/App/init_procedures.py
+++ b/Server/App/init_procedures.py
@@ -1,0 +1,33 @@
+import logging
+import subprocess
+
+import MySQLdb
+
+from config import DBConfig
+
+"""
+Simple script that initializes the DB with stored procedures.
+"""
+
+
+db = MySQLdb.connect(host=DBConfig.HOST, user=DBConfig.USERNAME, passwd=DBConfig.PASSWORD)
+
+if not db:
+    logging.error('Could not connect to DB')
+    exit(1)
+
+try:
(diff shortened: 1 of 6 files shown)


ClaytonPassmore:
@henryc132 this should make it convenient to reload stored procedures when we update them.. just do `make procedures`

Server/App/init_procedures.py
```diff
@@ -0,0 +1,33 @@
...
+db = MySQLdb.connect(host=DBConfig.HOST, user=DBConfig.USERNAME, passwd=DBConfig.PASSWORD)
+
+if not db:
+    logging.error('Could not connect to DB')
+    exit(1)
+
+try:
+    # Delete stored procedures in the Cue database.
+    cursor = db.cursor()
+    cursor.execute('DELETE FROM mysql.proc WHERE db LIKE \'{}\''.format(DBConfig.DB_NAME))
```
henryc132: Do we want to add type=procedure, or just delete stored functions in this step as well

ClaytonPassmore:
@henryc132 good call! Just deleting procedures now!

henryc132 (approved):

PR #52, opened 2017-03-07
Use mysql full text search for search endpoint

### LINKED ISSUE #51: Simplify search

henryc132:
Fixes #51 

@ClaytonPassmore 

Changed search procedure to take advantage of mysql's full text searching.

Decided to leave the Tags table in, even after adding the tags_delimited column since a delimited column is useless for any queries that isn't full text searching

### DIFF
--- a/Server/App/api/discover.py
+++ b/Server/App/api/discover.py
@@ -4,11 +4,11 @@ import logging
 from config import StatusCode
 from utils.serializer import serializer
 from utils.base_handler import BaseHandler
-from utils.wrappers import authorize_request_and_create_db_connector, require_query_string_params
+from utils.wrappers import authorize_request_and_create_db_connector, require_query_string_params, optional_query_string_params
 
 class TopHandler(BaseHandler):
     @authorize_request_and_create_db_connector
-    @require_query_string_params('page')
+    @optional_query_string_params('page')
     def get(self, page, connector):
         """Retrieve metadata for the most popular decks."""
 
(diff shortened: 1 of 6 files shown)


ClaytonPassmore:
@henryc132 are we going to have any other queries on tags that are going to require that tags table? At this point we aren't doing any categorization or anything like that.

Server/App/api/discover.py
```diff
@@ -42,15 +42,12 @@ def get(self, page, connector):
...
+    @optional_query_string_params('page')
+    def get(self, page, query, connector):
         """Retrieve metadata for decks matching the search string."""
 
-        tags = json.loads(tags_json)
-
-        tags_string = ",".join(tags)
-
-        decks = connector.call_procedure('SEARCH_DECKS', query, tags_string, len(tags), page)
+        decks = connector.call_procedure('SEARCH_DECKS', query, page)
```
ClaytonPassmore, line 50: For all of these `page` optional params, we're going to have to do a `None` check. If it's `None`, we're going to have to set it to 0.

ClaytonPassmore (changes requested):
We should perhaps consider eliminating the `Tags` table.. Let me know your thoughts!

Server/App/api/discover.py
henryc132: I actually had that check in the stored procedure

ClaytonPassmore: Yup I totally missed that, nice!

ClaytonPassmore (approved):

ClaytonPassmore:
@henry we can always remove the tags table later. :ship:

henryc132:
I mainly thought of keeping tags in case we want to use it in the future for some sort of advance search or similar feature.

ClaytonPassmore:
@henryc132 I feel like we probably will have other priorities, but whatever, I'm not opposed to keeping it in for now just in case.

ClaytonPassmore:
GTM :motor_boat:

PR #53, opened 2017-03-07
Make target for populating DB with fake data

ClaytonPassmore:
Run `make testdata` to redo the schema and stored procedures, plus add some test data.
The test data should be useful as we get more features up and running.

I made the `TEST_DATA_FILES` an array so that we can add more test data later in separate files as needed.


henryc132 (approved):
Do those user's fb_user_id actually link to a real account?

ClaytonPassmore:
@henryc132 yup but they're useless without passwords/fb access tokens

PR #57, opened 2017-03-08
Refine Deck View, add filter support, add common EmptyView

mattyu007:
Deck View v2!

## Summary

1. `DeckView` is now split into `DeckView.ios`, `DeckView.android`, and `CardListView`, and the layouts have been tweaked slightly.
  * Likewise, `DeckViewInfoHeader` has also split into `.ios` and `.android`.
  * iOS play button has moved into the bottom toolbar, card count has been removed, icons have been tweaked.
2. Filtering in `DeckView` is now functional
3. New common `EmptyView` base view for when there are no items to display

## Implementation Notes
In `DeckView.android`, we need to listen to the layout of the header and scroll events from the the main scroll view since we need to dynamically set and unset the `elevation` on the native toolbar.

### Empty States
This icon sucks and the layout is kind of wonky since `flex: 1` does nothing inside a `ScrollView`, but it's a start.

empty

If a deck is filtered and there aren't any flagged cards to show, we display an empty message. Notice the toggled state of the filter button in the bottom-left on iOS and in the top toolbar on Android.

empty_flagged

### Standard Deck

A standard deck.

regular

When the deck is scrolled, the header (title block) scrolls under the native toolbar and disappears. Notice that on Android, the list has extra padding at the bottom to avoid having content permanently obscured by the FAB.

scroll

### Shared Deck

shared

### Public Deck

public


mattyu007:
You know, the more I look at it, the more I want to change the iOS header to be more in line with the Android header with the tinted background and the tag-like Shared / Public indication... Thoughts?

Tyler-Brenneman (approved):

Tyler-Brenneman:
Merging so I can use react-native-material

PR #71, opened 2017-03-08
refactor api classes to use common CueApi class

Tyler-Brenneman:
Refactored the api classes to use a common class called CueApi

The CueAPi has 1 static method `fetch(endpoint: string, method: string = 'GET', body = null)`

I stored the userId and token as a static property in the class. This static property is set during:
LOGGED_OUT action,
LOGGED_IN action,
State persistence (app start)

This should solve the problem of storing the auth header. I tried using the `connect` method but it wasn't working for me.

Not sure if this is react-native best practice, but it worked in the testing I did.

### DIFF
--- a/Cue/app/api/CueApi.js
+++ b/Cue/app/api/CueApi.js
@@ -1,0 +1,46 @@
+// @flow
+'use strict'
+
+import { serverURL } from '../env';
+
+class CueApi {
+	static userId  = null;
+	static accessToken = null;
+
+	static setAuthHeader(userId: ?string, accessToken: ?string) {
+		this.userId = userId;
+		this.accessToken = accessToken;
+	}
+
+	static fetch (endpoint: string, method: string = 'GET', body: ?string = null): Promise<JSON> {
+		let headers = null;
+		if (this.userId != null && this.accessToken != null) {
+			headers = {
+				'X-CUE-USER-ID' : this.userId,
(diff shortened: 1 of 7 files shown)


Cue/app/api/CueApi.js
```diff
@@ -0,0 +1,46 @@
...
+		.then(checkStatus)
+		.then(response => response.json())
+		.catch(e => {
+			throw (e)
+		});
+	}
+}
+
+var checkStatus = function(response) {
+	if (response.status >= 200 && response.status < 300) {
```
mattyu007: This check can actually be replaced with `if (response.ok) {`

mattyu007 (approved):
Looks good. I'm not sure this is the most Redux-y way to do it either (since I think that would be somehow grabbing the state object inside each Action), but this is simple enough and makes sense.

Cue/app/api/CueApi.js
Tyler-Brenneman: The auth api returns created(201) on first time authenticating a user, and a 200 response every time after

mattyu007: Looks like `.ok` returns `true` if the response code is between 200 and 299, but this is tiny tiny nitpick: https://developer.mozilla.org/en-US/docs/Web/API/Response/ok

Tyler-Brenneman: OK cool, i was looking at that wrong. I made that change and rebased

Issue #45, opened 2017-03-02
State isn't cleared after logout


Tyler-Brenneman:
The redux state is not entirely cleared when logging out.
This can lead to users potentially being able to see other user's library's if they login on same device and library synch fail. The previous user's library state will not be reset and will be visible as if it was your own.

PR #72, opened 2017-03-08
set initial state on logged_out action

Tyler-Brenneman:
Should fix issue #45 
Will need to handle the LOGGED_OUT case for every reducer class we add

### DIFF
--- a/Cue/app/reducers/index.js
+++ b/Cue/app/reducers/index.js
@@ -4,8 +4,17 @@ import user from './user';
 import tabs from './tabs';
 import library from './library';
 
-module.exports = combineReducers({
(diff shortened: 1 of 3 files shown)


mattyu007:
Would it be better to create a root reducer which acts on `LOGGED_OUT` instead (like [this SO question](http://stackoverflow.com/questions/35622588/how-to-reset-the-state-of-a-redux-store)) so we don't have to manage the action in every reducer we create?

Tyler-Brenneman:
That is a much more elegant solution. I pushed a commit with this approach.

mattyu007 (approved):

PR #73, opened 2017-03-09
Fetch by share code and model overhaul

### LINKED ISSUE #68: Create endpoint for fetch deck by share code
Should be similar to the fetch deck by uuid endpoint.

### LINKED ISSUE #69: Add deck endpoint should take optional share code param
If an add deck request for a non-public deck is accompanied by a valid share code for the requested deck, the deck should be added to the user's library

ClaytonPassmore:
- Can now fetch deck via share code
- Created models that interact with the DB so handlers don't have to
- Created policies for accessing models and moved that logic out of the stored procedures
- Simplified deck retrieval so that we don't have a bunch of similar stored procedures
- Created migration mechanism for DB changes
- Added a linter
- Removed DeckTag and Tag tables
- Updated stored procedures to use Deck.tags_delimited
- Created mechanism for nested transactions
- Misc code cleanup

Closes #68 
Closes #69


Server/App/api/library.py
```diff
@@ -1,56 +1,38 @@
...
-
-            connector.call_procedure('LIBRARY_ADD', user_id, uuid)
-
-            connector.end_transaction()
+        deck = DeckPolicy.can_view(uuid, user_id, share_code, connector=connector)
+        if deck is None:
+            return self.make_response(status=StatusCode.NOT_FOUND)
 
+        try:
+            connector.call_procedure_transactionally('LIBRARY_ADD', user_id, deck.uuid)
```
henryc132: Might as well make this a function of the Library model to keep things consistent?

Server/App/api/library.py
```diff
@@ -113,8 +76,11 @@ class LibraryDeckRemoveHandler(BaseHandler):
     def delete(self, user_id, connector, uuid):
         """Remove a deck from a user's library and delete it if they are the owner."""
 
-        connector.call_procedure('LIBRARY_REMOVE', user_id, uuid)
-
+        try:
+            connector.call_procedure_transactionally('LIBRARY_REMOVE', user_id, uuid)
```
henryc132: Same as library_add

Server/App/model/discover.py
```diff
@@ -0,0 +1,42 @@
...
+class Discover(object):
+    def __init__(self, user_id, connector=None):
+        self.user_id = user_id
+        self.connector = connector or DatabaseConnector()
+
+    def fetch(self, category, page=1):
+        deck_uuids = self.connector.call_procedure('FETCH_DECK_UUIDS', category, page)
+        results = []
+        for deck_uuid in deck_uuids:
+            results.append(Deck(deck_uuid['uuid'], connector=self.connector).get_metadata(self.user_id))
```
henryc132, line 17: Seems like a lot of individual requests for decks in the new scheme, instead of just getting them in mass through joins. Seems like this could get rather inefficient with larger numbers?

ClaytonPassmore: I totally hear what you're saying but my theory was that with tags, we had to do an extra request for each individual deck to retrieve the tags anyway. So I figured its still the same number of queries. Also, it allows us to only have to maintain the DECK_FETCH procedure, so now we don't have a bunch of code duplication on the SQL side, which I'm happy with, because maintaining stored procedures sucks.

Server/App/policy/deck.py
```diff
@@ -0,0 +1,41 @@
...
+            return None
+
+        if metadata['owner'] != user_id:
+            return None
+        return deck
+
+    @staticmethod
+    def in_library(deck_id, user_id, connector=None):
+        user = User(user_id, connector=connector)
+        if deck_id in map(lambda x: x['deck_id'], user.get_deck_uuids_in_library()):
```
henryc132: Seems overkill to grab ids of all decks in library just to perform a check? I feel like it would be better for that logic to be in a procedure

Server/App/utils/database.py
```diff
@@ -25,22 +25,33 @@ def __del__(self):
...
         self.connector.rollback()
         self.connector.autocommit(True)
 
     def call_procedure_transactionally(self, proc_name, *params):
         """Call a stored procedure in the database and JSON-ify the results."""
-        self.begin_transaction()
-        result = self.call_procedure(proc_name, *params)
-        self.end_transaction()
+        try:
+            self.begin_transaction()
```
henryc132, line 49: is wrapping call_procedure in transactions necessary?

henryc132: Also the alternating use of call_procedure_transactionally and call_procedure gets confusing haha

Server/App/model/deck.py
```diff
@@ -0,0 +1,127 @@
...
+            return result[0]['share_code']
+
+        # Need to generate a share code.
+        self.connector.begin_transaction()
+        while True:
+            share_code = ''.join(random.choice(string.ascii_uppercase + string.digits)
+                                 for _ in range(SecurityConfig.SHARE_CODE_LENGTH))
+
+            result = self.connector.call_procedure('SET_SHARE_CODE', self.uuid, share_code)
+            if len(result) != 1:
```
henryc132: Looks like this If will stop the loop if the generated share code was a duplicate, but we want to regenerate a new share code and attempt to add it.

Server/App/utils/database.py
ClaytonPassmore: Yeah I tend to wrap the procedures that write in a transaction. If the procedure is read only then I don't usually wrap it in a transaction. Also, if there are multiple procedures that need to get called to modify something, I wrap them in one bigger transaction.

Server/App/utils/database.py
```diff
@@ -78,7 +89,11 @@ def query(self, query, *params):
         return results
 
     def query_transactionally(self, query, *params):
-        self.begin_transaction()
-        results = self.query(query, *params)
-        self.end_transaction()
+        try:
+            self.begin_transaction()
```
henryc132, line 93: Also same thing here, I don't think there is any reason to wrap single queries in transactions

Server/SQL/procedures.sql
```diff
@@ -31,47 +32,42 @@ END$$
 
 
 -- Retrieve the name of a user with their UUID.
-CREATE PROCEDURE GET_USER_NAME(IN u VARCHAR(36))
+CREATE PROCEDURE GET_USER(IN u VARCHAR(36))
 BEGIN
-    SELECT name FROM User WHERE uuid=u;
+    SELECT * FROM User WHERE uuid=u;
```
henryc132: Might as well just list out the fields?

Server/App/utils/database.py
ClaytonPassmore: If we have a write query that affects multiple entries it should be done transactionally

ClaytonPassmore:
@henryc132 ready for another look!

henryc132 (approved):

PR #82, opened 2017-03-12
cueApi static header should now be set properly on app restarts

Tyler-Brenneman:
Should fix bug for `CueApi` static auth properties not being set properly on app restarts.

This bug happened because I was trying to set the `CueApi` headers directly after `persistStore()`, which happens async. This caused the `CueApi` headers to be set with `null` before the actual store was loaded from `persistStore(...)`

### DIFF
--- a/Cue/app/setup.js
+++ b/Cue/app/setup.js
@@ -4,6 +4,7 @@ var CueApp = require('./CueApp');
 var React = require('React');
 var { Provider } = require('react-redux');
 var configureStore = require('./store/configureStore');
+import CueApi from './api/CueApi'
 const FBSDK = require('react-native-fbsdk');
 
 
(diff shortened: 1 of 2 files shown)


Cue/app/setup.js
```diff
@@ -29,6 +30,7 @@ function setup(): ReactClass<{}> {
       if (this.state.isLoading) {
         return null;
       }
+      CueApi.setAuthHeader(this.state.store.getState().user.userId, this.state.store.getState().user.accessToken);
```
mattyu007: Why not do this in the `configureStore` callback?

mattyu007: Actually this is ok too, since we'll only ever reach this line if `isLoading` is false, meaning the callback has been called.

Tyler-Brenneman: I tried that at first, but this.store is undefined during the callback. Not sure why, the debugger says they are defined and I can access them, but react-native throws a `Cannot read property 'getState' of undefined`

Tyler-Brenneman: Wowww k im dumb, my debugger was lying to me and had a different context. I will push a commit using callback method

henryc132 (approved):
Looks good

Issue #46, opened 2017-03-02
Library view doesn't refresh when LOADED_LIBRARY action is received


mattyu007:
After login, the library and deck data is downloaded async, which means a stale Library screen is displayed while the sync is happening in the background. The Library screen needs to refresh when the `LOADED_LIBRARY` action is fired.

PR #83, opened 2017-03-12
refresh library view when decks state changes

Tyler-Brenneman:
Refreshes the `LibraryListView` datasource whenever new props are passed in.

Also made a change to retrieve all deck's cards async so the loading will be faster.

This closes issue #46 

Right now there is currently no indication in the library view that the library is being loaded. It just says the library has no decks. I think while it is being loaded, a loading spinner could be displayed instead of the empty library view

### DIFF
--- a/Cue/app/tabs/library/LibraryListView.js
+++ b/Cue/app/tabs/library/LibraryListView.js
@@ -60,7 +60,7 @@ class LibraryListView extends React.Component {
       data[section].push(deck)
     }
 
-    this.props.decks.forEach((deck) => {
+    decks.forEach((deck) => {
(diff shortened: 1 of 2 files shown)


Cue/app/tabs/library/LibraryListView.js
```diff
@@ -60,7 +60,7 @@ class LibraryListView extends React.Component {
       data[section].push(deck)
     }
 
-    this.props.decks.forEach((deck) => {
```
mattyu007, line 63: Oops, nice catch.

mattyu007 (approved):

PR #85, opened 2017-03-12
Refresh Library View

### LINKED ISSUE #77: Add support for loading indicator and pull to refresh from Library view

ClaytonPassmore:
- Added a refresh control to the library list view
- Updated deck categorization to use `accession` attribute

Closes #77

### DIFF
--- a/Cue/app/tabs/library/LibraryListView.js
+++ b/Cue/app/tabs/library/LibraryListView.js
@@ -3,12 +3,11 @@
 'use strict';
 
 import React from 'react'
-import { Navigator, View, Text, Platform, ListView, Dimensions } from 'react-native'
-
-import { connect } from 'react-redux'
+import { Navigator, View, Text, Platform, ListView, Dimensions, RefreshControl } from 'react-native'
 
 import type { Deck, Card } from '../../api/types'
 
+import CueColors from '../../common/CueColors'
 import CueIcons from '../../common/CueIcons'
 import EmptyView from '../../common/EmptyView'
 import ListViewHeader from '../../common/ListViewHeader'
@@ -30,9 +29,8 @@ const SECTION_PUBLIC = 'Public'
 type Props = {
   navigator: Navigator,
   decks: Array<Deck>,
-
-  // filled in by Redux
-  userId: string
+  refreshing: boolean,
+  onSwipeToRefresh: () => void
 }
 
 class LibraryListView extends React.Component {
@@ -52,7 +50,7 @@ class LibraryListView extends React.Component {
   }
 
   _categorizeDecks(decks: Array<Deck>) {
-    let data = { }
+    let data = {}
     let addToData = (section, deck) => {
       if (!data[section]) {
         data[section] = []
@@ -61,9 +59,9 @@ class LibraryListView extends React.Component {
     }
 
     decks.forEach((deck) => {
-      if (deck.owner === this.props.userId) {
+      if (deck.accession === 'private') {
         addToData(SECTION_PRIVATE, deck)
-      } else if (!deck.public && deck.share_code) {
+      } else if (deck.accession === 'shared') {
         addToData(SECTION_SHARED, deck)
       } else {
         addToData(SECTION_PUBLIC, deck)
@@ -100,7 +98,6 @@ class LibraryListView extends React.Component {
     }
   }
 
-
   componentWillReceiveProps(newProps: Props) {
     this.setState({
       ...this.state,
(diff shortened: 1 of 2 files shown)


Cue/app/actions/library.js
```diff
@@ -17,4 +17,16 @@ async function loadLibrary(): PromiseAction {
   };
 }
 
-module.exports = { loadLibrary };
+async function setLibrary(library): PromiseAction {
```
mattyu007: Doesn't this do the same thing as `loadLibrary`?

Cue/app/tabs/library/LibraryListView.js
```diff
@@ -83,24 +86,39 @@ class LibraryListView extends React.Component {
     return { data, headers }
   }
 
+  _onRefresh() {
```
mattyu007: You can declare this using the new arrow syntax to avoid needing to call `this._onRefresh.bind(this)` in the `RefreshControl`, i.e., `_onRefresh = () => {...`

Cue/app/tabs/library/LibraryListView.js
```diff
@@ -137,8 +161,9 @@ class LibraryListView extends React.Component {
 
 function select(store) {
   return {
-    userId: store.user.userId
-  }
+    userId: store.user.userId,
```
mattyu007: Since we're using `accession` to categorize decks, we don't actually need to pull `userId` from the store anymore.

Cue/app/tabs/library/LibraryListView.js
```diff
@@ -137,8 +161,9 @@ class LibraryListView extends React.Component {
...
+    decks: store.library.decks
```
mattyu007: I feel like it might be better to have `LibraryListView` not connected to the Redux store at all, but have its state solely controlled by `refreshing: boolean`, `onSwipeToRefresh: () => void`, and `decks: Array<Deck>` props passed in by `LibraryHome`, which would be the one to actually call the API in `onSwipeToRefresh` and have the `decks` prop update propagated down to the list view automatically by React.

Cue/app/tabs/library/LibraryListView.js
ClaytonPassmore: Cool! Didn't know you could do that with a class method

Cue/app/actions/library.js
ClaytonPassmore: Uhm its really similar... I just realized I should just dispatch `loadLibrary`, bear with me :wink:

Cue/app/tabs/library/LibraryListView.js
ClaytonPassmore: Yes! This sounds a lot better

ClaytonPassmore:
@mattyu007 updated! I think this makes a lot more sense.

One thing I'm unsure of: I still connected the `LibraryListView` to the Redux store simply so I could have the `dispatch` function in the `props`. Is there a better way?

mattyu007:
@ClaytonPassmore Yeah, you can push the connection to Redux up to `LibraryHome` (since it's already connected to the store anyway to allow Create New Deck to work) by adding two extra props to `LibraryListView`:

1. `refreshing: boolean`, to show whether `LibraryHome` is currently waiting for `loadLibrary()` to return
2. a `onSwipeToRefresh: () => void` callback to get `LibraryHome` to call `loadLibrary` and set `refreshing` to `true`

I'm not sure if this is over-"normalizing", but I think the fewer pieces we have interacting with Redux, the better. Plus, if we end up needing to disallow entering Edit Mode or creating a new deck while the library is refreshing, this enables us to do that.

ClaytonPassmore:
@mattyu007 updated again! Let me know if this is more along the lines of what you were thinking!

Cue/app/tabs/library/LibraryListView.js
```diff
@@ -129,16 +125,16 @@ class LibraryListView extends React.Component {
         dataSource={this.state.dataSource}
         initialListSize={8}
         pageSize={2}
+        refreshControl={
+          <RefreshControl
+            refreshing={this.props.refreshing}
```
mattyu007: Nitpick: Can we add `colors={[CueColors.primaryTint]}` to the `RefreshControl` to tint the inner loading circle on Android?

Cue/app/tabs/library/LibraryListView.js
```diff
@@ -129,16 +125,16 @@ class LibraryListView extends React.Component {
...
+            onRefresh={this.props.onSwipeToRefresh}
+          />
+        }
         renderSectionHeader={(decks, category) => <ListViewHeader section={category} />}
         renderRow={deck => <LibraryListViewItem navigator={this.props.navigator} deck={deck} />} />
-      )
+      );
```
mattyu007: Nitpick: Semicolons aren't required in JS in most cases.

mattyu007 (approved):

Cue/app/tabs/library/LibraryListView.js
ClaytonPassmore: True, but why are we using `'use strict';` then?

mattyu007: Uhh, habit? 😬

PR #86, opened 2017-03-13
Disable editing for non-private decks

### LINKED ISSUE #70: Disable editing features on non-owned decks
Mostly in Deck View. Editing features should be disabled / removed and replaced with a Copy option.

ClaytonPassmore:
- Added new copy icon
- If deck is not private, take away editing options
  - On Android, show copy icon
  - On iOS, show "Copy" text

**iOS is untested.**

Closes #70


Cue/app/common/CueIcons.js
```diff
@@ -16,6 +16,7 @@ module.exports = {
   overflow: require('./img/overflow.png'),
   plus: require('./img/plus.png'),
   share: require('./img/share.png'),
+  copy: require('./img/copy.png'),
```
mattyu007: Can we keep this in alphabetical order? 😝

Cue/app/tabs/library/DeckView.android.js
```diff
@@ -98,28 +98,39 @@ class DeckView extends React.Component {
...
     ]
 
+    if (this.props.deck.accession === 'private') {
+      rightItems = rightItems.concat([
+        {
+          title: 'Add card',
+          icon: CueIcons.plus
+        },
+        {
+          title: 'Deck sharing options',
```
mattyu007: Should we restrict the sharing options to only owned decks? I'd imagine that we could still show the Deck Sharing Options screen, but it would all be read only. (So that people could look up the share codes and stuff.)

Cue/app/tabs/library/DeckView.ios.js
```diff
@@ -54,17 +54,26 @@ class DeckView extends React.Component {
...
-        title: 'Add Card',
-        icon: CueIcons.plus,
-        display: 'icon'
-      },
-      {
-        title: 'Edit',
-        display: 'text',
-      }
-    ]
+    let rightItems = []
```
mattyu007: You can just say `let rightItems` here and then assign it inside the `if`/`else` blocks without needing to re-assign or mutate.

Cue/app/tabs/library/DeckView.android.js
ClaytonPassmore: Should we do it based on whether or not there is a share code available for the deck? Because if there isn't a share code generated, should we be allowing someone that isn't the user to be generating a share code?

Cue/app/tabs/library/DeckView.ios.js
ClaytonPassmore: Good call!

Cue/app/tabs/library/DeckView.android.js
mattyu007: Mmm, I mean that the Deck Sharing Options screen can still be opened, but it would be display only and can't be edited. So if the deck is Shared, it would only display the sharing code, and if the deck is Public, then it would only tell you that the deck is public.

ClaytonPassmore: I could see it being useful for shared decks then, but not for public decks (unless a particular public deck already has a share code generated). Do we want to do it for all decks just to keep it consistent?

mattyu007: I think it's better to keep it consistent and leave the sharing options accessible in all cases, but selectively disable functionality based on the sharing status of the current deck. Plus, keeping it always accessible also always keeps the play button in the iOS toolbar centre-aligned in all cases. 😜

ClaytonPassmore: iOS QA over here... :wink:

ClaytonPassmore:
@mattyu007 updated! Not sure if I like the way I did it... style suggestions are welcome!

mattyu007 (approved):
Looks fine to me.

PR #88, opened 2017-03-13
Create deck + sync data structure

### LINKED ISSUE #59: Add Create New Deck flow
When the + button on the Library page is tapped, the user should enter the Create New Deck flow.

UX Reference:
create_new_deck

Tyler-Brenneman:
closes #59

Added a new object to library store. `localChanges` This object stores an array of objects of changes to a deck. I think it will be easiest to show how it works through these state changes:

image
image

Note that when a deck is created, not all the metadata is available because it is created by the server. After the sync for a deck happens the metadata will be updated. Not sure if not having this metadata will be a problem for any other features.

Right now the deck will only be synced with the server when doing a library refresh.

Split `LibraryHome.js` Into 2 different files, `LibraryHome.ios.js` and `LibraryHome.android.js`

Also added support for `DECK_DELETED` and `DECK_EDITED` actions

## New packages:
- `react-native-device-info` - get device info (needed for some api calls)
- `react-native-prompt-android` - prompt for android that can enter text. Only needed for android because ios has `AlertIOS`. The only way I could find how to change the color of the buttons was to set styling in `styles.xml` in the android project
- `uuid` - needed to create a temp uuid for locally created decks(will pronably be needed for cards aswell). Once the deck is synced with server, the server uuid will be used instead

## Remaining:
- Sync only happens when refreshing library view
   - need to make issue for syncing on logout
- Doesn't reorder cards on position change
   - need to make issue for this

image


ClaytonPassmore:
@lonebutterfish in terms of metadata provided by the server, we may need a placeholder or two. I think having the generated `uuid` is good but we should also set the `owner` to be the `uuid` of the logged in user, and we should also set the accession to be `private`. (I might be missing an attribute or two).

Also, I think some of your edits have tabs instead of spaces. Might wanna change your editor settings :stuck_out_tongue:

Tyler-Brenneman:
@ClaytonPassmore I wasn't sure exactly how to get the userId in the library action when creating a new deck. Maybe I could just get it from CueApi? It would work but not sure it is the best practice

Does that approach sound reasonable or is there a better method to getting the userId?

ClaytonPassmore:
@lonebutterfish we have it in the Redux store already, so maybe we could get the component that is creating the deck to pass it as a parameter? I'm not sure if that's the _best_ way but I think an extra request is too much.

Tyler-Brenneman:
@ClaytonPassmore It wouldn't be an api request getting the userId from CueApi since the userId is just a static property I could access.

But I think ill go with passing it through as a parameter, think it would be better practice that way.

EDIT: going to go with the CueApi method. was having problems with binding the context so i gave up and went the easier way

ClaytonPassmore:
@lonebutterfish ah ok sorry, I misunderstood :stuck_out_tongue_closed_eyes:

Cue/app/tabs/library/LibraryHome.ios.js
```diff
@@ -0,0 +1,125 @@
...
+      }
+    }]
+
+    return (
+      <View style={styles.container}>
+      <CueHeader
+        leftItem={leftItem}
+        title='Library'
+        rightItems={rightItems}
+        overflowItems={overflowItems} />
```
mattyu007: `overflowItems` isn't defined anywhere and isn't used on iOS anyway.

Cue/app/tabs/library/LibraryHome.ios.js
```diff
@@ -0,0 +1,125 @@
...
+          'Create New Deck',
+          'Enter a name for the new deck.',
+          [
+           {text: 'Cancel', style: 'cancel'},
+           {text: 'Create', onPress: (deckName) => {
+             let deck = this.props.onCreateDeck(deckName).deck
+             this.props.navigator.push({deck})
+           }},
+          ],
+          {
```
mattyu007: The `type` and `placeholder` are actually just extra params to `AlertIOS.prompt`, so I wasn't able to get this to work until I moved them to become params to `prompt`.

Cue/app/reducers/library.js
```diff
@@ -1,26 +1,73 @@
...
+        return decks[i] = serverDeck
+      }
+    })
+    localChanges.find( (deck, i) => {
+      if (deck.uuid = change.uuid)
+        return localChanges.splice(i,1)
+    })
+  }
+
+  return {
```
mattyu007: I think you're running into this because you're changing the state even if the action isn't handled by this reducer, so the `persist` actions are being clobbered. Adding an extra case to the end of the `if`/`else if`  blocks before the `return` statement fixed the persistence issue for me:
```js
  else {
    return state
  }
```

Cue/app/actions/library.js
```diff
@@ -17,4 +19,80 @@ async function loadLibrary(): PromiseAction {
...
+    owner: CueApi.userId,
+    accession: 'private',
+  }
+  return {
+    type: 'DECK_CREATED',
+    deck,
+  }
+}
+
+function editDeck(deck: DeckChanges) : Action {
```
mattyu007: What is `DeckChanges`?

Cue/app/actions/library.js
```diff
@@ -17,4 +19,80 @@ async function loadLibrary(): PromiseAction {
...
+    })
+    Promise.all(promises).then(() => {
+      if (failedSyncs.length == 0) dispatch(loadLibrary())
+    })
+    console.log('failed syncs:  ', failedSyncs)
+    return failedSyncs
+  };
+}
+
+// deck should only contain changes in the deck to avoid sending excess data
```
mattyu007: The `library` reducer doesn't do this diffing for edited cards right now, right?

Cue/app/api/Library.js
```diff
@@ -1,19 +1,58 @@
...
+      cards: cards,
+    })
+		console.info('Creating deck "' + name + '" at endpoint ' + endpoint)
+    return CueApi.fetch(endpoint, 'POST', body)
+  },
+
+  editDeck (change){
+    let endpoint = 'api/v1/deck/' + change.uuid
+    let body = JSON.stringify({
+      device: DeviceInfo.getUniqueID(),
```
mattyu007: Should we have a user-facing "friendly name" for devices for the purposes of making merge conflicts easier on users? @ClaytonPassmore @henryc132

Cue/app/actions/library.js
Tyler-Brenneman: Originally I tried to create some flow types for the localChanges, but flow hates me and I couldn't get it working without a million errors. The `DeckChanges` is remnants of that. Ill remove it.

Cue/app/api/Library.js
Tyler-Brenneman: I can use `getModel()` instead. Will return something like "iPhone 7"

Cue/app/reducers/library.js
Tyler-Brenneman: yup this fixed, it. Thanks.

Cue/app/actions/library.js
Tyler-Brenneman: It should be able to handle it, but i havn't tested any of the edit deck action... The idea is that the library reducer should be able to handle it

Cue/app/tabs/library/LibraryHome.ios.js
```diff
@@ -82,10 +82,8 @@ class LibraryHome extends React.Component {
              this.props.navigator.push({deck})
            }},
           ],
-          {
-            type: 'plain-text',
-            placeholder: 'My Great Deck',
-          },
+          type: 'plain-text',
```
mattyu007: This still doesn't work, `'plain-text'` and `'My Great Deck'` have to be just normal parameters.

Cue/app/reducers/library.js
```diff
@@ -1,26 +1,75 @@
...
+    })
+    if (!localChanges.find( (localChange, i) => {
+      if (localChange.uuid == uuid && localChange.action == 'add')
+        return localChanges.splice(i,1)
+    })) {
+      localChanges.push({uuid, action: 'delete'})
+    }
+
+  } else if (action.type == 'DECK_EDITED') {
+    let editedDeck = action.deck
```
mattyu007: Should this be `let editDeck = action.*change*` to match up with the change to the library `editDeck` action?

Cue/app/reducers/library.js
```diff
@@ -39,16 +39,16 @@ function library(state: State = initialState, action: Action): State {
...
-      if (deck.uuid == editedDeck.uuid)
-        return decks[i] = { ...deck, editedDeck }
+      if (deck.uuid == change.uuid)
+        return decks[i] = { ...deck, change }
     })
     if (!localChanges.find( (deck, i) => {
-      if (deck.uuid == editedDeck.uuid)
-        return localChanges[i] = { ...deck, editedDeck }
+      if (deck.uuid == change.uuid)
+        return localChanges[i] = { ...deck, change, action: deck.action }
```
mattyu007: Sorry for all the back and forth... What happens if the `action` in `localChanges` is an `add` action? If it gets replaced with an `edit` action, wouldn't that cause problems when we try to sync? I think we'll need to do some smarter merging in the future to handle these cases.

Tyler-Brenneman: `return localChanges[i] = { ...deck, change, action: deck.action }` should keep it as an add action if it originally was an add action. This way the edits are also in the add action.

When `syncDeck` action gets called it should see its an add action, create the deck, then check if there are any other changes that need to be synced. If there is, then call the editApi and sync the remaining edits.
The `syncDeck` action doesnt support this yet, but i put it as a todo

Tyler-Brenneman: Not sure if that syntax is correct though. Should it be `return localChanges[i] = { ...deck, ...change, action: deck.action }` ?

Tyler-Brenneman: Same as line 45, should it be `return decks[i] = { ...deck, ...change }` ?

Cue/app/actions/library.js
```diff
@@ -17,4 +19,80 @@ async function loadLibrary(): PromiseAction {
...
+// ex: { uuid: string, action 'add', name: string } if a deck was added
+// ex: { uuid: string, action 'edit', cards: [{action: 'edit', card-id: string, front: "fox"}] } if 1 card in deck was changed
```
ClaytonPassmore: I think it's actually `card_id` and also the card would have to include the `back` as well, so:
```
{ action: 'edit', card_id: string, front: 'fox', back: '20XX' }
```

Cue/app/actions/library.js
```diff
@@ -17,4 +19,80 @@ async function loadLibrary(): PromiseAction {
...
+async function syncDeck(change) : PromiseAction {
+  let serverDeck
+  try {
+    if (change.action == "add"){
```
ClaytonPassmore: nitpick: `===` instead of `==` on these

Cue/app/api/Library.js
```diff
@@ -1,19 +1,58 @@
...
+    })
+		console.info('Creating deck "' + name + '" at endpoint ' + endpoint)
+    return CueApi.fetch(endpoint, 'POST', body)
+  },
+
+  editDeck (change){
+    let endpoint = 'api/v1/deck/' + change.uuid
+    let body = JSON.stringify({
+      device: DeviceInfo.getDeviceName(),
+			parent_deck_version: change.parent_deck_version,
```
ClaytonPassmore: This line and the next couple are tabs instead of spaces

Cue/app/api/Library.js
```diff
@@ -1,19 +1,58 @@
...
+			tags: change.tags,
+			actions: change.cards //TODO: api expects card_id, not uuid
+    })
+    console.info('editing deck' + body + ' at endpoint ' + endpoint)
+    return CueApi.fetch(endpoint, 'PUT', body)
+  },
+
+  deleteDeck(uuid: string) {
+    let endpoint = 'api/v1/library/' + uuid
+		console.info('deleteing deck at endpoint ' + endpoint)
```
ClaytonPassmore: Another tab here

Cue/app/reducers/library.js
```diff
@@ -1,28 +1,127 @@
...
+
+  } else if (action.type === 'DECK_CREATED') {
+    decks.push(action.deck)
+    localChanges.push({...action.deck, action: "add"})
+
+  } else if (action.type === 'DECK_DELETED') {
+    let deckIndex = decks.findIndex(deck => deck.uuid == action.uuid)
+    if (deckIndex >= 0)
+      decks.splice(deckIndex,1)
+    let changeIndex = localChanges.findIndex(deck => deck.uuid == action.uuid)
```
ClaytonPassmore: Is it possible that we could have multiple changes for the same deck in `localChanges`? If so, this would only find the first change I think

Cue/app/reducers/library.js
```diff
@@ -1,28 +1,127 @@
...
+    if (changeIndex >= 0) {
+      if (localChanges[changeIndex].action === 'add')
```
ClaytonPassmore: Should we also be checking for edit actions?

Cue/app/reducers/library.js
```diff
@@ -1,28 +1,127 @@
...
+    } else {
+      localChanges.push({...change, action: 'edit'})
+    }
+
+  } else if (action.type === 'DECK_SYNCED') {
+    let change = action.change
+    let serverDeck = action.serverDeck
+    let deckIndex = decks.findIndex(deck => deck.uuid == change.uuid || deck.uuid == serverDeck.uuid)
+    if (deckIndex >= 0)
+      decks[deckIndex] = serverDeck
```
ClaytonPassmore: Should we log some kind of error if we don't find a matching deck here?

Cue/app/reducers/library.js
```diff
@@ -1,28 +1,127 @@
...
+    let changeIndex = localChanges.findIndex(deck => deck.uuid == change.uuid)
+    if (changeIndex >= 0)
+      localChanges.splice(changeIndex,1)
```
ClaytonPassmore: Could there be a scenario where we had an edit and also a delete, so we perform the edit and then remove the delete from the local changes? Or will the edit/delete be collapsed down into a single entry in the local changes?

Cue/app/reducers/library.js
```diff
@@ -1,28 +1,127 @@
...
+
+  return {
+    decks,
+    localChanges
   }
 }
 
+//merges card edits for a deck's cards
+function _mergeCards(cards, cardsChanged) {
+  if (!cards) return []
```
ClaytonPassmore: In this scenario, should we return cardsChanged?

Cue/app/reducers/library.js
```diff
@@ -1,28 +1,127 @@
...
+  if (!cardsChanged) return cards
+  let mergedCards = cards.slice();
+  cardsChanged.forEach(cardChange => {
+    let index = mergedCards.findIndex(card => card.uuid == cardChange.uuid)
```
ClaytonPassmore: For edits, this might end up being `card_id` rather than `uuid`

Cue/app/reducers/library.js
```diff
@@ -1,28 +1,127 @@
...
+  }
+
+  return {
+    decks,
+    localChanges
   }
 }
 
+//merges card edits for a deck's cards
+function _mergeCards(cards, cardsChanged) {
```
ClaytonPassmore: I'm a little worried about the relative ordering of cards here. Say we have an initial edit that moves card A to position 2, and then we have another edit that moves card A to position 1 and move Card B to position 2.. Something like that.. we need to make sure we apply those operations in the correct order... It would almost be worth it to just straight up concatenate the two arrays together so that all operations get applied in the order they were intended to.

Cue/app/reducers/library.js
```diff
@@ -1,28 +1,127 @@
...
+  if (!cards) return cardsChanged
+  if (!cardsChanged) return cards
+  let mergedCards = cards.slice();
+  cardsChanged.forEach(cardChange => {
+    let index = mergedCards.findIndex(card => card.uuid == cardChange.uuid)
+    if (index >= 0) {
+      if (cardChange.action === 'delete' && cards[index].action === 'add')
+        mergedCards.splice(index,1)
+      else {
+        mergedCards[index] = {...mergedCards[index], ...cardChange, action: cards[index].action}
```
ClaytonPassmore: I'm just worried again here about relative card order..

Cue/app/tabs/library/LibraryHome.ios.js
```diff
@@ -0,0 +1,122 @@
...
+      refreshing: false
+    });
+  }
+
+  constructor(props: Props) {
+   super(props);
+   this.state = {
+     refreshing: false
+   };
+   this._refresh = this._refresh.bind(this)
```
ClaytonPassmore: I'm not sure that you need to bind this.

Cue/app/reducers/library.js
Tyler-Brenneman: localChanges should only contain 1 entry for every deck that is changed. If multiple edits happen to 1 deck it should aggregate all the edits into 1 object

Cue/app/reducers/library.js
Tyler-Brenneman: So this checks if the localChange is an add action, if it is then that means that deck was only created locally and isn't in the server yet. We can delete that localChange and won't need to sync anything. If it is an edit action, that means the deck is already in the server and we need to add a localChange to delete the deck from the server.

Cue/app/api/Library.js
Tyler-Brenneman: ahhh why is my IDE so weird. ill fix this

Cue/app/reducers/library.js
Tyler-Brenneman: If a deck was deleted then we don't need to sync what edits happened to it, so the localChange is updated to be a delete action. (see line 38)

Cue/app/reducers/library.js
Tyler-Brenneman: I kept the key as `uuid` so that everything uses same key and makes merging simpler between `decks` and `localChanges`. In the api call we should convert `uuid` to `card_id` where needed

Cue/app/reducers/library.js
Tyler-Brenneman: we won't find a matching deck if change was a delete action, since in the `'DECK_DELETED'` action would have already deleted it from `decks`

Cue/app/reducers/library.js
Tyler-Brenneman: this scenario should really never happen. Only if you somehow try to sync your deck before your library is loaded after login. Ill update it though to handle it better.

Cue/app/tabs/library/LibraryHome.ios.js
Tyler-Brenneman: I get errors about `this` being undefined in `_refresh` without it

Cue/app/reducers/library.js
Tyler-Brenneman: Yea this needs fixing.
We need to update all of the `decks.cards`'s position so that they display in the correct order locally until a sync happens.

For `local_changes` we can just concatenate the two arrays and let the server handle that merge.

Cue/app/tabs/library/LibraryHome.ios.js
mattyu007: Might be because `_refresh` is only declared as an arrow function in `LibraryHome.ios`, not in `LibraryHome.android`.

Cue/app/reducers/library.js
mattyu007: Ah, yes. You would need to unpack them both, otherwise it would just be the same as writing `{...deck, change: change}`. I believe this only does a *shallow merge* though, so only the latest set of changes will be picked up.

You might need to manually merge the two `deck.cards` and `change.cards` arrays. If you just concatenate them both in order (i.e., `{...deck, ...change, cards: [...deck.cards, ...change.cards]}`, I think that should also work as long as the server is able to process successive edits to a single card in a single request.

Cue/app/tabs/library/LibraryHome.android.js
```diff
@@ -0,0 +1,160 @@
...
+        key: 'Select',
+        title: 'Select'
+    }]
+
+    const AddFAB = MKButton.coloredFab()
+      .withBackgroundColor(CueColors.primaryTint)
+      .withStyle(styles.fab)
+      .withOnPress(() => {
+        prompt(
+          'Create New Deck',
```
mattyu007: Extremely minor nitpick: According to Android / Material conventions, this should be in sentence case, i.e., `Create new deck`. This is wrong in the UX mock, sorry.

Tyler-Brenneman: updated

Cue/app/tabs/library/LibraryHome.ios.js
Tyler-Brenneman: It's because of the arrow function. Fixed and removed the bind.

Cue/app/actions/library.js
```diff
@@ -17,4 +19,89 @@ async function loadLibrary(): PromiseAction {
...
+  return (dispatch, getState) => {
+    let failedSyncs = []
+    let promises = []
+    localChanges.forEach(change =>{
+      promises.push(dispatch(syncDeck(change))
+        .catch(e => {
+          failedSyncs.push(change)
+        }))
+    })
+    return Promise.all(promises).then(() => {
```
mattyu007, line 62: I just looked this up, and apparently `Promise.all` actually rejects immediately if any of the sub-promises rejects. So if I'm reading this correctly, `failedSyncs` will only ever have 0 or 1 item in it, even if multiple syncs failed. Perhaps take a look at http://stackoverflow.com/questions/31424561/wait-until-all-es6-promises-complete-even-rejected-promises ?

mattyu007 (commented):
Looking good! Just a minor thing with the behaviour of `Promise.all` which we might need to look into.

Cue/app/actions/library.js
mattyu007: Actually on second look, we already catch any sub-promises which reject. So this won't be an issue?

ClaytonPassmore:
One thing is that right now if a deck gets deleted by another user (ie a remote deck), we are still pulling the changes down which will nuke our local copy of the deck and we will lose all of the cards. There needs to be some control method injected in here somewhere (maybe in a different PR?)

Cue/app/actions/library.js
Tyler-Brenneman: Yea my hope was the sub promise catch would be able to handle it,  but i never actually tested it so im not sure. Looking at the stack overflow discussion it looks like catching the sub promises should work.

Tyler-Brenneman:
Yea i think that should go in different PR. I wont have time to address that till this weekend .

How do we want to handle this case? Warn the user that a remote deck has been deleted and ask them if they want to make a local copy?

mattyu007:
~~That~~ A warning + copy prompt would be good, but I think we should create a separate issue for that warning flow.

mattyu007 (approved):

PR #89, opened 2017-03-13
Deck Insets

### LINKED ISSUE #62: Add deck insets to Library view for shared and public decks owned by the user
In Library view, decks which are owned by the user and which are currently shared / public should have a coloured inset in the top-right corner.

(Note that even though it is in the UX reference, at this time we are not supporting drag and drop to rearrange decks.)

UX Reference:
cue ux mocks 005

ClaytonPassmore:
- Added deck insets
- Updated test data

Closes #62 

@mattyu007 do we want a diagonal border across the hypotenuse of the inset?

insets

### DIFF
--- a/Cue/app/tabs/library/LibraryListViewItem.js
+++ b/Cue/app/tabs/library/LibraryListViewItem.js
@@ -3,11 +3,12 @@
 'use strict';
 
 import React from 'react'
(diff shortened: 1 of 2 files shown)


mattyu007 (approved):
I think it's fine without a border along the hypotenuse. The border is supposed to surround the entire card, and the inset is part of the card.

PR #90, opened 2017-03-14
Add Play Deck and setup interstitial

### LINKED ISSUE #61: Add Play Deck screens
The ability to play decks.

Here's a sample tutorial on how to implement drag & drop in RN: http://moduscreate.com/animated_drag_and_drop_with_react_native/

**Note this also includes the interstitial setup screen where users can select whether they want to shuffle the deck and/or where in the deck they would like to start.**

UX Reference:
cue ux mocks 014
cue ux mocks 015
cue ux mocks 016
cue ux mocks 017

mattyu007:
This PR adds the Play Deck view and setup interstitial, closes #61, and handles our pivot.

**This is a huge PR;** take your time. 😉 

## Summary

1. Upgraded to React Native 0.42 — This means we can now specify element dimensions as percentages (e.g., `width: '80%'`).
  * Note that you may need to install Java 8 if you are having issues building the Android project.
2. Added `PlayDeckSetupView` — This is an interstitial screen where users will be able to select how they want to play a deck (ordered, shuffled, or from a specific card).
3. Added `PlayDeckView` — The main attraction. Screenshots are below, but please try it out on an emulator or a device.

## TODO & Outstanding Questions

1. The flag card swipe gesture doesn't do anything right now.
2. Shadows on Android don't work right now. `elevation` doesn't play well with `react-native-flip-card`, so if we need to fix this, we might just need to write our own flipping card component.
3. Should `PlayDeckView` have some kind of indication as to whether the current deck is being shuffled or not?
4. In landscape and on devices of different sizes, the floating gesture action indicators don't come to rest at the exact centre of the screen. Is this ok? See screenshots at the end.

## Screenshots

When the play button in the deck view is tapped, the `PlayDeckSetupView` interstitial screen appears. Users are able to select between playing the deck in sequential order, random order, or sequential starting at a specific card.

If the user chooses to begin at a specific card, a picker appears to allow the selection to occur.

After tapping "Play" or the forward icon, the `PlayDeckView` view appears.

**All interactions are the same, so these GIFs just show the iOS Simulator.**

Cards default to showing the front. Tapping a card causes it to flip over and show the back

Begin swiping to show the action indicator / hints. Release to hide.

Swipe left to continue to the next card. Swiping left after the last card will pop the navigation stack.

Swipe up to flag or unflag the current card and move to the next card.

Swipe right to go back.

Swipe right is disabled when the current card is the first card.

Swipe down from the header to close.

Or tap the down arrow to close.

Of course, everything also works in landscape, barring a few minor layout issues.
landscape


Cue/app/tabs/library/play/PlayDeckSetupView.js
```diff
@@ -0,0 +1,140 @@
...
+              ? this.state.customStartIndex
+              : undefined
+          })}
+      }
+    ]
+    return (
+      <View style={{flex: 1}}>
+        <CueHeader
+          leftItem={leftItem}
+          title={'Play “' + this.props.deck.name + '”'}
```
ClaytonPassmore, line 131: I'm not sure if I like having the deck title in quotes.

Cue/app/tabs/library/play/PlayDeckView.js
```diff
@@ -0,0 +1,487 @@
...
+
+  componentWillUnmount() {
+    StatusBar.setHidden(false)
+  }
+
+  _shuffled(array: Array<*>): Array<*> {
+    let ret = array.slice(0)
+
+    for (let i = ret.length - 1; i > 0; i--) {
+      let j = Math.floor(Math.random() * (i + 1))
```
ClaytonPassmore, line 145: This was cleverly constructed :1st_place_medal:

Cue/app/tabs/library/play/PlayDeckView.js
```diff
@@ -0,0 +1,487 @@
...
+        toValue: toValue,
+        easing: Easing.ease,
+        duration: DURATION_STANDARD,
+      }),
+      this._showOnlyIndicator(action)
+    ])
+  }
+
+  _commitFlagAction = () => {
+    // TODO Dispatch a Redux action?
```
ClaytonPassmore, line 250: :+1:

Cue/app/tabs/library/play/PlayDeckView.js
```diff
@@ -0,0 +1,487 @@
...
+      Animated.timing(this.state.flagOpacity, flagConfig),
+      Animated.timing(this.state.nextOpacity, nextConfig),
+      Animated.timing(this.state.backOpacity, backConfig)
+    ])
+  }
+
+  _setUpIndicatorAnimation = () => {
+    Animated.spring(
+      this.state.flagY, {
+        toValue: this.state.cardXY.y.interpolate({
```
ClaytonPassmore, line 414: What is this doing exactly?

ClaytonPassmore:
@mattyu007 AFAICT this looks good! It looks amazing on the Android emulator!
I got a little lost in some of the animation code TBH, but from what I do understand, it makes sense.
Once you resolve your conflicts, I'm good :+1:

Cue/app/tabs/library/play/PlayDeckSetupView.js
mattyu007: Yeah they don't look great especially when truncated, but since this title string mixes user-generated content with app "template" text, I wanted a way to keep the two distinct. Or are we ok with just leaving them out?

Cue/app/tabs/library/play/PlayDeckView.js
mattyu007: The classic Durstenfeld Shuffle 😉 Definitely didn't Google for that one.

Cue/app/tabs/library/play/PlayDeckView.js
mattyu007: Mmhmm, waiting on #88.

Cue/app/tabs/library/play/PlayDeckView.js
mattyu007: For the flag indicator, we animate its movement along the Y axis by having its `bottom` style value use a spring animation to track the y-value of the card's position. (The position of the card itself is controlled using `this.state.cardXY` which is updated by the `PanResponder` attached to the card view — see `_onPanResponderMove` and `_onPanResponderRelease`.) Since a 1:1 correspondence between them would mean that the indicator could be permanently obscured by the card, we use `interpolate` to reduce the tracking to 0.4:1 so the indicator "lags" behind the card as the card moves towards the top of the screen. The `inputRange` and `outputRange` are set up like so:

```
              ---*---
inputRange:  [ -1, 0, 1]
outputRange: [0.4, 0, 0]
                  --**--
```
\* Since the y-value of the card decreases as the card is moved upwards, but we want the `bottom` value to increase, we map a decrease of 1 in the card y-value to an increase of 0.4 in the `bottom` style value.
\** This range is used for extrapolation so that if the card moves down, the `bottom` value never decreases below 0, ensuring that the indicator never slides off the screen.

Hope that made sense! A similar strategy is used for the next and back indicators, which track the x-value of `this.state.cardXY`  instead.

Cue/app/tabs/library/play/PlayDeckSetupView.js
ClaytonPassmore: Yeah I understand that approach. I wouldn't mind seeing how it looks without the quotes, but I don't feel too strongly about it

Cue/app/tabs/library/play/PlayDeckView.js
ClaytonPassmore: Ah ok yes! This makes sense!

ClaytonPassmore (approved):

PR #93, opened 2017-03-14
Allow empty library refresh

### LINKED ISSUE #91: Can't refresh library if library is empty
@ClaytonPassmore 

found this while creating new decks

ClaytonPassmore:
- Add dummy `ListView` in front of `EmptyView` as per @mattyu007's suggestion
  - Allows the library to be refreshed even when empty

**Needs iOS testing.**

Fixes #91

### DIFF
--- a/Cue/app/tabs/library/LibraryListView.js
+++ b/Cue/app/tabs/library/LibraryListView.js
@@ -3,7 +3,7 @@
 'use strict';
 
 import React from 'react'
-import { Navigator, View, Text, Platform, ListView, Dimensions, RefreshControl } from 'react-native'
+import { Navigator, View, Text, Platform, ListView, Dimensions, RefreshControl, StyleSheet } from 'react-native'
 
 import type { Deck, Card } from '../../api/types'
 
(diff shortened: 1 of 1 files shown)


Cue/app/tabs/library/LibraryListView.js
```diff
@@ -106,16 +106,32 @@ class LibraryListView extends React.Component {
...
-        <EmptyView
-          icon={CueIcons.emptyLibrary}
-          titleText={'You don&rsquo;t have any decks yet.'}
-          subtitleText={Platform.OS === 'android'
-            ? 'Create one or add one to your Library from Discover in the menu.'
-            : 'Create one or add one to your Library from the Discover tab.'} />
+        <View style={{flex: 1}}>
+          <EmptyView
+            icon={CueIcons.emptyLibrary}
+            titleText={'You don&rsquo;t have any decks yet.'}
```
mattyu007: While you're here, can you also replace this escape with the literal curly quote character: `’`?

Cue/app/tabs/library/LibraryListView.js
```diff
@@ -106,16 +106,32 @@ class LibraryListView extends React.Component {
...
+            subtitleText={Platform.OS === 'android'
+              ? 'Create one or add one to your Library from Discover in the menu.'
+              : 'Create one or add one to your Library from the Discover tab.'} />
+          <ListView
+            style={{...StyleSheet.absoluteFillObject}}
```
mattyu007: You don't need to unpack and re-pack the `absoluteFillObject` here.

mattyu007 (approved):
Seems to work fine on iOS. A few minor nitpicks, but looks good otherwise.

Cue/app/tabs/library/LibraryListView.js
mattyu007: Actually you should be able to use `StyleSheet.absoluteFill` directly as the style

PR #97, opened 2017-03-16
Add Card Entry view

### LINKED ISSUE #60: Add Create New Card / Edit Card flow
When the + button in Deck View is tapped, or a card in Deck View is tapped while in Edit Mode, the user enters the Create New Card / Edit Card flow. I'm considering these two items to be the same flow since the user experience is essentially the same, but implementation details may differ.

UX Reference:
create_edit_card

mattyu007:
Closes #60. To help make it a little easier for us to test. 😜

## A note about the most latest commit
The latest commit (bce8e99) makes two major changes:

1. Combines `DeckView.ios` and `DeckView.android` back into a single view.
    * To avoid copy-paste code for connecting to the Redux store and firing off actions
    * This makes layout a little hairy since the iOS and Android variants of this screen vary quite a bit. However, since most of the components are the same (but are just laid out in different areas), a new `_getAllCommonItems` method simply generates all the necessary items, and then the various rendering methods (`_getLeftItems`, `_getRightItems`, `_renderToolbar`, etc) simply pick out the items they need. **Please let me know if you think this is a bad idea and if the view should remain split into two files.**
2. Connects `DeckView` to the Redux store.
    * Since the `CardListView` subcomponent now needs to re-render live when changes are made to the store (i.e., new cards are added or existing cards are edited), the `DeckView` must be directly connected to Redux store since it has no parent component to receive updated props from.
    * This also means that `DeckView` no longer receives `deck: Deck` prop, but instead receives `deckUuid: string` from the Navigator route and `decks: Array<Deck>` from Redux, and has to perform the deck lookup itself.

## Summary
This PR adds a card entry view, supporting both the create card and edit card flows. Note that at this time, the `onSubmit` callback isn't hooked into the Redux actions yet.

### New Card
The "New Card" mode is triggered when the route is pushed with a `null` value for `cardEntry`.
screen shot 2017-03-16 at 2 40 18 pm
screen shot 2017-03-16 at 2 42 17 pm

The "Edit Card" mode is triggered when a non-`null` value is pushed.
screen shot 2017-03-16 at 2 43 45 pm


Cue/app/tabs/library/DeckView.ios.js
```diff
@@ -60,7 +60,19 @@ class DeckView extends React.Component {
...
           title: 'Add Card',
           icon: CueIcons.plus,
-          display: 'icon'
+          display: 'icon',
+          onPress: () => {
+            this.props.navigator.push({
+              cardEntry: null,
+              onSubmit: (front: string, back: string, existingUuid: ?string) => {
+                // TODO
+                alert('Received card input:'
```
ClaytonPassmore: Is this just until we hook it into the sync mechanism?

mattyu007: Yes, since #88 is in now, I'll replace this.

Cue/app/common/TextEntryTableRow.js
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  // We need both onChange and onContentSizeChange for height since:
+  //  - onContentSizeChange only fires on initial mount on Android, instead of
+  //    whenever the content size changes (as expected, like on iOS)
+  //  - onChange doesn't fire on initial mount so we can't rely on it to set
+  //    the initial size of the text box
+  _onChange = ({nativeEvent: {contentSize: {height}, text}}) => {
+    this.setState({
+      text,
```
ClaytonPassmore, line 60: how is this an object without a key?

mattyu007: It's a [destructuring assignment](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Pulling_fields_from_objects_passed_as_function_parameter) for extracting items from a function parameter. Functionally it's equivalent to:
```js
_onChange = (event) => {
  let height = event.nativeEvent.contentSize.height
  let text = event.nativeEvent.text
  // ...
}

**Comment by ClaytonPassmore** [comment]
help

bad_card_add

**Comment by mattyu007** [comment]
Oh yeah, there's a bug in the library reducer, which should be merged in with #100 (commit 3d3005e).

File: Cue/app/common/TextEntryTableRow.js (hunk 2 above)
Comment (ClaytonPassmore) [reply Cue/app/common/TextEntryTableRow.js:60]: Ah ok, this is fancy. I haven't seen this before

File: Cue/app/tabs/library/DeckView.js [hunk 3]
```diff
@@ -0,0 +1,293 @@
...
+      return dispatch(editDeck(change))
+    },
+    editCard: (deckUuid: string, cardUuid: string, front: string, back: string) => {
+      let change = {
+        uuid: deckUuid,
+        cards: [
+          {
+            action: 'edit',
+            uuid: cardUuid,
+            front,
```
ClaytonPassmore, line 283: this looks trippy but I'm slowly getting over it

ClaytonPassmore (approved):

PR #100, opened 2017-03-16
Add support for actually flagging decks in Play Deck View

### LINKED ISSUE #96: Actually flag cards in Play Deck when cards are swiped up
Right now the swipe up gesture just moves to the next card without firing off any actions.

Also consider indicating on the card itself whether it is flagged or not.

mattyu007:
The big 1-0-0! Closes #96. 😀

## Summary
`PlayDeckView` now takes a snapshot of the state of the deck when it is presented. Since it is not connected to any component which is actively receiving updates from Redux, `PlayDeckView` will dispatch `editDeck` actions while simultaneously updating its local snapshot.

## Issues
Since we call `setState` before the indicators disappear, the indicators can briefly display the state intended for the next card. You can see this at the end of the GIF below when the `REMOVE FLAG` indicator briefly turns into a `FLAG FOR REVIEW` indicator before disappearing. Not sure if it's a high priority to fix.

## GIF
flag

### DIFF
--- a/Cue/app/tabs/library/play/PlayDeckView.js
+++ b/Cue/app/tabs/library/play/PlayDeckView.js
@@ -3,6 +3,9 @@
 import React from 'react'
 import { Navigator, Animated, Easing, View, Text, Image, TouchableOpacity, PanResponder, Platform, Dimensions, StatusBar } from 'react-native'
 
+import { connect } from 'react-redux'
+import { editDeck } from '../../../actions'
+
 import type { Deck, Card } from '../../../api/types'
 
 import CueColors from '../../../common/CueColors'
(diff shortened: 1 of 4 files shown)


Cue/app/tabs/library/play/PlayDeckView.js
```diff
@@ -97,7 +103,8 @@ export default class PlayDeckView extends React.Component {
   constructor(props: Props) {
     super(props)
 
-    let cards = this.props.deck.cards || []
+    // Take a snapshot of the Deck in props
+    let cards = (this.props.deck.cards || []).slice(0)
```
ClaytonPassmore, line 107: Pretty sure slice defaults to 0 but nbd

ClaytonPassmore (approved):

PR #108, opened 2017-03-17
Re-add deck thumbnail insets, soften some corners, fix search list item text on iOS

mattyu007:
## Summary
This PR:

* re-adds insets to `DeckThumbnail` which were missed during the refactor of `DeckThumbnail`.
* softens the look of `DeckThumbnail` the Library and Discover home pages.
* fixes the `SearchListViewItem` text not appearing on iOS.

## Implementation Notes
1. In addition to `itemContainer` having `borderRadius: 2`, the insets themselves also have `borderRadius: 2` applied to them since `overflow: hidden` is not supported on Android when used in combination with `borderRadius`, and `overflow: hidden` cuts off the shadow on iOS. The inset assets have had extra whitespace added to protect the left and bottom corners from becoming inadvertently rounded.
    * Side note: We really need better inset assets than just the letters `S` and `P` in triangles...
2. Since it's possible for a deck in Discover Home to already been in the user's library, `DeckThumbnail` now has an extra optional `hideInsets` prop to prevent insets from being shown.

### Library
1-library-before
2-library-after

### Discover
3-discover-before
4-discover-after

### Search
5-search-before
6-search-after


Cue/app/common/DeckThumbnail.js
```diff
@@ -57,6 +86,22 @@ export default class DeckThumbnail extends React.Component {
       // Display "Updating..."?
     }
 
+    let inset
+    if (!this.props.hideInsets) {
+      if (this.props.deck.accession === 'private') {
+        if (this.props.deck.public) {
```
henryc132, line 92: Shouldn't we show the public inset regardless of whether or not the user owns the deck?

mattyu007: Hmm, I don't think so, because it doesn't add any detail. Public decks in the user's library are already sorted under the "Public" heading so it would just be redundant information.

Cue/app/tabs/library/LibraryListView.js
```diff
@@ -148,7 +148,7 @@ class LibraryListView extends React.Component {
         initialListSize={8}
         pageSize={2}
         refreshControl={refreshControl}
-        renderSectionHeader={(decks, category) => <ListViewHeader section={category} />}
+        renderSectionHeader={(decks, category) => <ListViewHeader style={{marginLeft: -16}} section={category} />}
```
ClaytonPassmore, line 151: dat outdent tho

ClaytonPassmore (approved):
This looks awesome!

Cue/app/tabs/library/LibraryListView.js
mattyu007: Yeah... It's not ideal, but it works. I wish margins collapsed in RN.

Issue #110, opened 2017-03-18
Edit Deck API route should support nulling out share_code


mattyu007:
According to the client, the categorization of decks with `accession: private` is as follows:
```js
if (deck.public) {
  // Deck is Public
} else if (deck.share_code) {
  // Deck is Shared with share_code
} else {
  // Deck is Private
}
```

To make a deck transition from Shared to Private, the `share_code` needs to be nulled out. The Edit Deck API currently ignores any `share_code` attributes sent from the client, which means it's not possible to transition a deck from Shared to Private (or from Public to Private if a deck previously had a `share_code` assigned).

PR #112, opened 2017-03-18
Add Deck Sharing Options screen

### LINKED ISSUE #67: Add Deck Sharing Options screen
Add the ability to manage deck sharing options.

Note the UX reference below is outdated. When tags are entered, as soon as a space, comma, or other punctuation mark is entered, the current text becomes a chip, and subsequent typing goes towards the next tag.

deck_sharing_options

mattyu007:
Closes #67.

## Summary
This pull request:

1. Adds the Deck Sharing Options screen for managing sharing and tagging.
2. Adds client-side Redux actions and API calls for managing share codes.
3. Fixes miscellaneous issues in library reducer, iOS `LibraryHome`, small assets.

## Issues you may run into while testing
- #110: Decks cannot be converted from Shared to Private.
- #114: After removing the last tag from a Public deck and syncing with the server, an empty string tag will appear.

## Implementation notes
* The share code generation API call happens inside the `DeckSharingOptions` component, and a `SHARE_CODE_GENERATED` Redux action is dispatched after the call returns to persist the share code to the Redux store. The share code action does not get enqueued to the `localChanges` queue since sync is not necessary.
* Changing the status of a deck happens by dispatching `DECK_EDITED` actions, meaning changes only take effect once the library is synced and local changes are pushed up the server.

## Screenshots
Basic private deck.
screen shot 2017-03-18 at 2 58 52 am

When anything in the view changes, the form is marked as dirty, and the header options change from Close to Cancel and Save. If there is no share code for the deck yet, the user is prompted to save the form to generate a share code.
screen shot 2017-03-18 at 3 01 14 am

Once the share code is generated, the form is marked as clean, and the code is displayed.
screen shot 2017-03-18 at 3 03 09 am

Share code generation needs to happen online.
screen shot 2017-03-18 at 3 05 10 am

For public decks, tags are displayed as chips.
screen shot 2017-03-18 at 3 08 17 am

Tap on the hint text, start typing, then press Done on the keyboard (or hit Enter) to create a tag. Notice that as soon as a tag is added, the form is marked as dirty. Tags are normalized before being saved. Tap on an X to remove a tag.
ios-chips
android-chips

If a deck is downgraded from Public to Private, a warning message appears.
screen shot 2017-03-18 at 3 14 19 am

A similar message appears for Shared to Private.
screen shot 2017-03-18 at 3 15 17 am

**Open question: Do we need a warning for Public to Shared?**

Sharing options are disabled for non-owned decks.
screen shot 2017-03-18 at 3 58 04 am
screen shot 2017-03-18 at 3 58 35 am


Cue/app/actions/types.js
```diff
@@ -12,6 +12,7 @@ export type Action =
   | { type: 'DECK_DELETED', uuid: string }
   | { type: 'DECK_SYNCED', serverDeck: Deck, change: {} }
   | { type: 'DECK_EDITED', change: {} }
+	| { type: 'SHARE_CODE_GENERATED', uuid: string, code: string}
```
ClaytonPassmore: Spacing messed up here?

Cue/app/tabs/library/DeckSharingOptions.js
```diff
@@ -0,0 +1,391 @@
...
+        message = 'The share code for your deck will become invalid immediately.'
+          + '\n\nPeople who currently have your deck in their Library will still be able to make copies of your deck.'
+        break
+    }
+
+    if (message) {
+      Alert.alert(
+        (Platform.OS === 'android'
+          ? 'Others may lose access to this deck'
+          : 'Others May Lose Access to this Deck'),
```
ClaytonPassmore: Is this really how the capitalization works on iOS? You obviously know better than I do

Cue/app/tabs/library/DeckSharingOptions.js
```diff
@@ -0,0 +1,391 @@
...
+    }
+
+    if (!this.props.deck.public) {
+      this.props.setPublic(this.props.deck.uuid, true)
+    }
+
+    this.props.navigator.pop()
+  }
+
+  _handleSharedSubmit = () => {
```
ClaytonPassmore, line 131: If we are going from public to shared, should we display some kind of warning message? Should we still let users that had the deck when it was public be able to access it if it moves to shared?

Cue/app/tabs/library/TagsTableRow.js
```diff
@@ -0,0 +1,91 @@
...
+}
+
+type Props = {
+  tags: Array<string>,
+  disabled?: boolean,
+  onTagAdded: (tag: string) => void,
+  onTagRemoved: (tag: string) => void,
+}
+
+export default class TagsTableRow extends React.Component {
```
ClaytonPassmore, line 32: Should this be `TagsInput`? Also, I think we need to cap out tags at a max of 500 chars (which is including commas delimiters).

ClaytonPassmore (commented):
@mattyu007: left you a couple of comments, let me know what you think.

Cue/app/actions/types.js
mattyu007: Switched to using spaces.

Cue/app/tabs/library/DeckSharingOptions.js
mattyu007: Yeah, alert titles are generally written in title case except in the rare case that they're formulated as complete sentences, in which case they're written using sentence case.

ClaytonPassmore: That's the attention to detail I like to see.

Cue/app/tabs/library/DeckSharingOptions.js
mattyu007: I'm not sure, it depends on what the actual behaviour would be... Do the decks disappear from Libraries? Are they automatically promoted to Shared? #117?

Cue/app/tabs/library/TagsTableRow.js
mattyu007: You mean the class name? It's named this way since it displays in the style of a table row, which is based loosely off the [grouped `UITableView` style in iOS](https://developer.apple.com/ios/human-interface-guidelines/ui-views/tables/).

I'll add the limit to the `_onTagAdded` callback in `DeckSharingOptions`.

Cue/app/tabs/library/DeckSharingOptions.js
ClaytonPassmore: So I think if the deck moves to shared, all users with `public` accession will lose the deck. If the deck was originally shared before it was public, users that have `shared` accession will still retain access to the deck. (I believe that's what we decided?)

Cue/app/tabs/library/TagsTableRow.js
ClaytonPassmore: Sounds good :+1:

mattyu007: screen shot 2017-03-18 at 10 46 21 pm

ClaytonPassmore: Perfect

mattyu007: Also... 500 characters is actually a LOT, I don't think anyone will ever run into it in practice.

ClaytonPassmore: LOL you know this because it took you 2 hours to type it all out

mattyu007: You know it.

ClaytonPassmore:
@mattyu007 looks good, only thing is if we should show a dialogue for making the deck shared when it was previously public... We could say something vague like "Many users will lose access to this deck, are you sure you want to proceed?" ... or something?

Cue/app/tabs/library/DeckSharingOptions.js
mattyu007: Hmm, ok. Would people who have `public` accession still be allowed to copy the deck before it's removed?

ClaytonPassmore: Yes

ClaytonPassmore: Only their local copy though.

mattyu007: OK I'm going to update the text for the 3 possibilities to such:

## Public -> Shared
### Others May Lose Access to this Deck
People who discovered your deck via Discover or Search will no longer receive updates for this deck.

`[Cancel]` `[Continue Anyway]`

## Public -> Private
### Others May Lose Access to this Deck
Your deck will no longer appear in Discover or Search.

People who currently have your deck in their Library will no longer receive updates for this deck.

`[Cancel]` `[Make Private Anyway]`

## Shared -> Private
### Others May Lose Access to this Deck

The current share code for your deck will be invalidated.

People who currently have your deck in their Library will no longer receive updates for this deck.

`[Cancel]` `[Make Private Anyway]`

ClaytonPassmore: I think this works. I think for the last two, we should maybe say "others will lose..." rather than "may"

ClaytonPassmore (approved):
Looks great! If you don't feel like changing "may" to "will" like I mentioned in my comment, that's fine. I don't feel too strongly about it.

Cue/app/tabs/library/DeckSharingOptions.js
mattyu007: Good point, *will* change.

ClaytonPassmore:
:boat: :motor_boat: :ship: :shipit:

mattyu007:
WOW I love GitHub Squash and Merge

PR #116, opened 2017-03-18
Misc. Server Fixes

### LINKED ISSUE #95: Add number of cards to standard metadata response
We shouldn't have to download an entire deck just to display the number of cards.

### LINKED ISSUE #107: Server returns 400 if needs_review is specified in add card action
I'm not sure if this is just a case of the API doc needing to be updated, but I thought I'd create an issue anyway.

When `needs_review` is passed in along with the `add` card deck edit action, the server returns a 400:
screen shot 2017-03-16 at 11 03 50 pm

Compare with the same card without `needs_review`:
screen shot 2017-03-16 at 11 07 11 pm

### LINKED ISSUE #110: Edit Deck API route should support nulling out share_code

### LINKED ISSUE #111: Special characters in device name cause server to return 400
The device name is sent along with Edit Deck API requests. If the device name contains any special characters, such as the curly quote in "Matthew’s MacBook Pro", then the server will return a 400.

Request from Android emulator which has device name "Unknown":
screen shot 2017-03-18 at 2 41 40 am

Identical request from iOS emulator which has device name "Matthew’s MacBook Pro":
screen shot 2017-03-18 at 2 43 22 am

iOS emulator request after changing computer name to just "MacBook Pro":
screen shot 2017-03-18 at 2 51 18 am

### LINKED ISSUE #114: Server returns an array of a single empty tag during deck edit when there are no tags
When all the tags are removed from a deck and the deck is synced, the server returns a deck with a single empty tag.

screen shot 2017-03-18 at 2 53 47 pm

### LINKED ISSUE #118: Deck overwrite api does not support needs_review flag
It's not mentioned in the api spec, but this endpoint should support the needs_review flag for cards.

image

ClaytonPassmore:
- Added `unshare` option to edit deck endpoint
- Allow unicode content in JSON
- Allow default handler methods from `BaseHandler` class to take any parameters (e.g. uuid='adsf')
- Allow tags to be set to `null`
- Fix bug where new cards could not be flagged
- Return number of cards in standard metadata deck response
- Allow cards to be marked for review with the overwrite endpoint

Closes #95
Closes #107
Closes #110
Closes #111
Closes #114
Closes #118

### DIFF
--- a/Server/App/api/deck.py
+++ b/Server/App/api/deck.py
@@ -56,8 +56,8 @@ class DeckUUIDHandler(BaseHandler):
     @authorize_request_and_create_db_connector
     @extract_user_id
     @require_params('parent_deck_version', 'parent_user_data_version', 'device')
-    @optional_params('name', 'public', 'tags', 'actions')
-    def put(self, name, public, tags, actions, parent_deck_version, parent_user_data_version, device,
+    @optional_params('name', 'public', 'unshare', 'tags', 'actions')
+    def put(self, name, public, unshare, tags, actions, parent_deck_version, parent_user_data_version, device,
             user_id, connector, uuid):
         """Edit an existing deck."""
 
@@ -85,6 +85,8 @@ class DeckUUIDHandler(BaseHandler):
                 deck.set_name(name)
             if public is not None:
                 deck.set_public(public)
+            if unshare is True:
+                deck.nullify_share_code()
             if tags is not None:
                 deck.set_tags(tags)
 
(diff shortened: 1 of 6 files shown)


Server/App/api/deck.py
```diff
@@ -85,15 +85,17 @@ def put(self, name, public, tags, actions, parent_deck_version, parent_user_data
                 deck.set_name(name)
             if public is not None:
                 deck.set_public(public)
+            if unshare is True:
+                deck.nullify_share_code()
```
henryc132, line 89: Looks like anyone who added the deck prior to it being unshared will still be able to fetch the newest versions of the deck. Is this intended?

Tyler-Brenneman:
I ran into another server issue just now with the overwrite endpoint. #118

Not sure if you want to get it with this PR as well, or if this PR fixes that too.

Server/App/api/deck.py
ClaytonPassmore: Yup. See #117

ClaytonPassmore:
@lonebutterfish hmm ok. I may try to get that in this PR. Thanks!

henryc132 (approved):

PR #120, opened 2017-03-19
Add accessible to metadata response

### LINKED ISSUE #117: Decide on process for when shared/public decks are made private
Currently if a deck becomes private, all the people that already had it in their library will be able to access it still. We need to think about the desired behaviour in this scenario.

Will need a server implementation as a followup.

ClaytonPassmore:
- `accessible` is a flag that indicates whether or not the user can still retrieve the given deck.

Closes #117.

**The bad news:** there is a bug right now so if a user had a shared deck, made it private, then made it shared again, users that added the deck the first time around via share code could potentially still have access to the deck because we don't save which share code you used to gain access to the deck.

### DIFF
--- a/Server/App/model/deck.py
+++ b/Server/App/model/deck.py
@@ -105,6 +105,7 @@ class Deck(object):
         if len(results) != 1:
             return None
         deck = results[0]
+        deck['accessible'] = self._is_accessible(deck, user_id)
 
         results = self.connector.call_procedure('GET_DELIMITED_TAGS', self.uuid)
         if len(results) != 1:
@@ -113,6 +114,15 @@ class Deck(object):
 
         deck['tags'] = [] if tags is None else tags.split(',')
         return deck
+
+    def _is_accessible(self, deck, user_id):
+        if deck['accession'] == 'private':
+            return True if user_id == deck['owner'] else False
+        if deck['accession'] == 'shared':
+            return True if (deck['public'] or deck['share_code'] is not None) else False
+        if deck['accession'] == 'public':
+            return True if deck['public'] else False
+        return False
 
     def get_full_deck(self, user_id):
         deck = self.get_metadata(user_id)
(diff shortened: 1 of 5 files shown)


ClaytonPassmore:
@henryc132 if you have any suggestions regarding the share code bug, I'm all ears.

Server/App/model/deck.py
```diff
@@ -114,6 +115,15 @@ def get_metadata(self, user_id):
         deck['tags'] = [] if tags is None else tags.split(',')
         return deck
 
+    def _is_accessible(self, deck, user_id):
+        if deck['accession'] == 'private':
+            return True if user_id == deck['owner'] else False
+        if deck['accession'] == 'shared':
+            return True if (deck['public'] or deck['share_code'] is not None) else False
+        if deck['accession'] == 'public':
+            return True
```
henryc132: Also need a deck['public'] == true right? Because it could no longer be public

Server/App/policy/deck.py
```diff
@@ -11,8 +11,7 @@ def can_view(deck_id, user_id, share_code=None, connector=None):
             return None
 
         if (
-            not metadata['public'] and
-            (user_id is None or metadata['owner'] != user_id) and
+            metadata['accessible'] is False and
             (share_code is None or metadata['share_code'] != share_code)
```
henryc132: probably no longer need this line if accessible is supposed to encapsulate all the conditions?

ClaytonPassmore: Accessible does not capture the scenario where a user hasn't added a shared deck to their library yet, but they have a share code for it.

Server/App/model/deck.py
ClaytonPassmore: Yes, thanks for catching that!

Server/App/policy/deck.py
ClaytonPassmore: `accessible` is going to only capture decks that are already in a user's library. I am going to add a check for public here as well.

henryc132:
In regards to the bug, we might as well just save the share code used in the Library table, along with accession?

henryc132 (approved):
Looks like theres another issue just for the bug, so I'll just approve this now

PR #122, opened 2017-03-19
Add extra bottom padding to LibraryListView on Android to avoid crashing into FAB

mattyu007:
Pretty self-explanatory.

Note that unlike in CardListView, we only need to account for the FAB's 16 pt of vertical padding once since the card containers already have 16 pt of bottom padding.

screen shot 2017-03-19 at 2 37 02 am


ClaytonPassmore (approved):
Sweet!

PR #124, opened 2017-03-19
Prevent empty decks from being played

### LINKED ISSUE #99: Error when playing empty deck
Error when playing a deck with 0 cards.

image

mattyu007:
Closes #99.

Display an alert instead of allowing an empty deck to be played.

screen shot 2017-03-19 at 3 11 15 am


ClaytonPassmore (approved):
lgtm

PR #125, opened 2017-03-19
Client fix for issue #95

### LINKED ISSUE #109: DeckMetadata type can't be spread into Deck type
It seems that Flow doesn't allow types to be spread using the `...` operator. This means that classes using the current compact `Deck` type won't type-check correctly using Flow.

screen shot 2017-03-17 at 8 13 53 pm

This should be a low-risk low-effort change, but we should look into perhaps using inheritance with Flow.

henryc132:
Quick fix to use the new num_cards field from server.

Closes #109

### DIFF
--- a/Cue/app/api/types.js
+++ b/Cue/app/api/types.js
@@ -14,6 +14,7 @@ export type DeckMetadata = {
   author: string,
   rating: number,
   num_ratings: number,
+  num_cards: number,
   tags: Array<string>,
   owner: string,
   public: boolean,
--- a/Cue/app/api/types.js
+++ b/Cue/app/api/types.js
@@ -28,8 +28,7 @@ export type DeckMetadata = {
   accession?: string,
 };
 
-export type Deck = {
-  ...DeckMetadata,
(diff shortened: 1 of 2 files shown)


Cue/app/api/types.js
```diff
@@ -14,6 +14,7 @@ export type DeckMetadata = {
   author: string,
   rating: number,
   num_ratings: number,
+  num_cards: number,
```
mattyu007, line 17: While you're in api/types.js, can you also take a look at #109? If you don't want to fiddle with inheritance in Flow, I'm ok with just copy/pasting DeckMetadata into Deck and adding a reminder comment to keep them synchronized.

henryc132: Oh man, never noticed it didnt actually work haha

henryc132:
@mattyu007 Fixed it with an intersection type. Doesn't look like there's any other way of extending a type.

mattyu007 (approved):
Looks good. Don't forget to squash and rebase before merging.

PR #128, opened 2017-03-20
Added test data for NFP validation

### LINKED ISSUE #127: Develop test case to validate NFP
We claimed that we should be able to handle 100 decks with 500 cards each.

Generate test data so that we can verify this claim.

ClaytonPassmore:
- Made a directory for test data files
- Wanted to buff my commit history so I'm adding this giant file (JK its so we can test our NFP of 100 decks x 500 cards)

So, I ran this on my android emulator. It ran out of memory and then I could still kind of use it, but then it crashed. So... can we still hit this NFP?

Closes #127


ClaytonPassmore:
@henryc132 on second thought... Maybe this isn't the best way of going about this...

PR #129, opened 2017-03-20
Add script to generate NFP test data

### LINKED ISSUE #127: Develop test case to validate NFP

ClaytonPassmore:
- Restructure test data layout
- New script

Closes #127


Server/Scripts/generate_nfp_testdata.py
```diff
@@ -0,0 +1,111 @@
...
+        else:
+            card += ','
+        print(card)
+
+print('\n')
+print('INSERT INTO `Library`')
+print('    (user_id, deck_id, version, last_update_device, accession)')
+print('VALUES')
+
+for i in xrange(100):
```
mattyu007: My Python is a little rusty, but couldn't you also do something like this to avoid needing to use the `if i !== 99` check?

```python
lib = ','.join(generate_lib(i) for i in xrange(100)) + ';'
```

Edit: Would also the overhead of constantly concatenating strings.

ClaytonPassmore: definitely could, then its just all in memory at the same time

ClaytonPassmore: Also I would like to point out that you wrote `!==`. You know you've been writing too much JavaScript when... :stuck_out_tongue_closed_eyes:

ClaytonPassmore:
Ping

Server/Scripts/init_test_data.py
```diff
@@ -0,0 +1,30 @@
...
+mysql_cmd = 'mysql -u{} -p'.format(DBConfig.USERNAME)
+
+try:
+    # Collect a list of test data files in order.
+    test_files = map(
+        lambda x: os.path.join(DBConfig.TESTDATA_DIRECTORY, x),
+        sorted(os.listdir(DBConfig.TESTDATA_DIRECTORY))
+    )
+    files = ' '.join(test_files)
+    mysql_cmd = 'mysql -u{} -p'.format(DBConfig.USERNAME)
```
mattyu007: `mysql_cmd` is defined twice?

ClaytonPassmore: I like to make sure its extra defined

ClaytonPassmore:
@mattyu007 updated!

mattyu007 (approved):

Issue #113, opened 2017-03-18
Use the flag cards endpoint instead of the edit deck endpoint when flagging cards


mattyu007:
Currently on the client when we flag cards, we call the edit deck endpoint, ~~which is kind of heavy-weight and causes the deck version to be incremented. We should be more efficient and call the flag cards endpoint instead, which only increments the user data version.~~

With further observation, this is actually a bug, since decks which have `shared` or `public` accession can't be edited using the edit deck endpoint, which makes flagging them this way impossible.

PR #149, opened 2017-03-23
Flag cards from the list view

### LINKED ISSUE #84: Add support for flagging cards in deck view
In Deck View, users should be able to tap on a card in the list to flag or unflag it.

ClaytonPassmore:
- Made CardListViewRow touchable if flagCard prop is defined
- Touching flags and unflags the card

moarflags

Closes #84
Still affected by #113

### DIFF
--- a/Cue/app/tabs/library/CardListView.js
+++ b/Cue/app/tabs/library/CardListView.js
@@ -14,7 +14,7 @@ type Props = {
   cards: Array<Card>,
   isFiltering: boolean,
   accession?: string,
-  flagCard?: (cardUuid: string, flag: boolean) => any,
+  onFlagCard?: (cardUuid: string, flag: boolean) => any,
 }
 
 export default class CardListView extends React.Component {
@@ -41,7 +41,10 @@ export default class CardListView extends React.Component {
 
   _getDataSource = (cards: Array<Card>, isFiltering: boolean) => {
     let ds = new ListView.DataSource({
-      rowHasChanged: (r1, r2) => r1.front !== r2.front || r1.back !== r2.back || r1.position !== r2.position
+      rowHasChanged: (r1: Card, r2: Card) => {
+        return (r1.front !== r2.front || r1.back !== r2.back || r1.uuid !== r2.uuid ||
+                r1.needs_review !== r2.needs_review)
+      }
     })
 
     cards = cards || []
@@ -81,7 +84,7 @@ export default class CardListView extends React.Component {
           ? { paddingBottom: 88 /* 56 + 2 x 16 to avoid FAB */ }
           : { } }
         dataSource={this.state.dataSource}
-        renderRow={card => <CardListViewRow card={card} flagCard={this.props.flagCard} />}
+        renderRow={card => <CardListViewRow card={card} onFlagCard={this.props.onFlagCard} />}
         renderSeparator={(section, row) => <ListViewHairlineSeparator key={row} />} />
     )
   }
--- a/Cue/app/tabs/library/CardListView.js
+++ b/Cue/app/tabs/library/CardListView.js
@@ -14,6 +14,7 @@ type Props = {
   cards: Array<Card>,
   isFiltering: boolean,
   accession?: string,
+  flagCard?: (cardUuid: string, flag: boolean) => any,
 }
 
 export default class CardListView extends React.Component {
@@ -40,7 +41,7 @@ export default class CardListView extends React.Component {
 
   _getDataSource = (cards: Array<Card>, isFiltering: boolean) => {
     let ds = new ListView.DataSource({
-      rowHasChanged: (r1, r2) => r1 !== r2
+      rowHasChanged: (r1, r2) => r1.front !== r2.front || r1.back !== r2.back || r1.position !== r2.position
     })
 
     cards = cards || []
@@ -80,7 +81,7 @@ export default class CardListView extends React.Component {
           ? { paddingBottom: 88 /* 56 + 2 x 16 to avoid FAB */ }
           : { } }
         dataSource={this.state.dataSource}
-        renderRow={card => <CardListViewRow card={card} />}
+        renderRow={card => <CardListViewRow card={card} flagCard={this.props.flagCard} />}
         renderSeparator={(section, row) => <ListViewHairlineSeparator key={row} />} />
     )
   }
--- a/Cue/app/tabs/library/CardListViewRow.js
+++ b/Cue/app/tabs/library/CardListViewRow.js
@@ -13,21 +13,25 @@ const styles = {
     paddingHorizontal: 16,
     paddingVertical: 12,
   },
+  textIconWrapper: {
+    flexDirection: 'row',
+    alignItems: 'flex-start',
+  },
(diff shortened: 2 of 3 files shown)


Cue/app/tabs/library/CardListView.js
```diff
@@ -39,7 +40,7 @@ export default class CardListView extends React.Component {
 
   _getDataSource = (cards: Array<Card>, isFiltering: boolean) => {
     let ds = new ListView.DataSource({
-      rowHasChanged: (r1, r2) => r1 !== r2
+      rowHasChanged: (r1, r2) => r1.front !== r2.front || r1.back !== r2.back || r1.position !== r2.position
```
ClaytonPassmore: Not really sure if this is helping

Cue/app/tabs/library/CardListView.js
```diff
@@ -13,6 +13,7 @@ import ListViewHairlineSeparator from '../../common/ListViewHairlineSeparator'
 type Props = {
   cards: Array<Card>,
   isFiltering: boolean,
+  flagCard?: (cardUuid: string, flag: boolean) => any,
```
ClaytonPassmore: Potentially undefined, so this is right, right?

Cue/app/tabs/library/CardListViewRow.js
```diff
@@ -21,24 +22,48 @@ const styles = {
...
+  _onPress = () => {
+    this.props.flagCard(this.props.card.uuid, !this.props.card.needs_review)
   }
 
   render() {
-    return (
+    let row = (
       <View style={styles.container}>
-        <Text style={styles.frontText}>
+        <Text style={{...styles.frontText, marginRight: this.props.card.needs_review ? 40 : 0}}>
```
ClaytonPassmore: Padding differs based on whether or not card is flagged

Cue/app/tabs/library/CardListViewRow.js
```diff
@@ -21,24 +22,48 @@ const styles = {
...
           {this.props.card.front}
         </Text>
         <Text style={styles.backText}>
           {this.props.card.back}
         </Text>
+        <Image
+          style={{...styles.flagIcon, opacity: this.props.card.needs_review ? 1 : 0}}
```
ClaytonPassmore: Flag is still there, just hide it by changing opacity if card is not flagged

Cue/app/tabs/library/CardListView.js
mattyu007: Yeah, if I annotate the type with `?` instead of the prop name and then don't specify the prop when creating the component, Flow will complain.

Cue/app/tabs/library/CardListView.js
mattyu007: I don't think it hurts, but I think the `position` check is superfluous, and I would also check if `uuid` or `needs_review` are different.

Feel free to also annotate the types of `r1` and `r2` to give your IDE some auto-complete magic.

Cue/app/tabs/library/CardListViewRow.js
```diff
@@ -21,24 +22,48 @@ const styles = {
...
+    position: 'absolute',
+    tintColor: CueColors.flagIndicatorTint
   }
 }
 
 export default class CardListViewRow extends React.Component {
   props: {
-    card: Card
+    card: Card,
+    flagCard: (cardUuid: string, flag: boolean) => any
```
mattyu007: This should also be marked as optional, and since this is a callback, I would call it `onFlagCard` instead.

Cue/app/tabs/library/CardListViewRow.js
mattyu007: You can specify multiple style objects using an array to avoid needing to spread `styles.frontText` to achieve the same thing, i.e., `style={[styles.frontText, {marginRight: ___}]}`

Cue/app/tabs/library/CardListViewRow.js
mattyu007: I'm not 100% convinced by this, I'd rather have the flag not rendered at all if the card isn't marked as needing review; see comment above.

Cue/app/tabs/library/CardListViewRow.js
mattyu007: Also, if you want to avoid manually adjusting the margin, you should also be able to use a wrapper view (around the title text and image) with `flexDirection: 'row', alignItems: 'flex-start'`. You'll need to set `flex: 1` on the text and `flex: 0` on the image, but this will also avoid needing to set `position: 'absolute'` on the image and allow the text to reflow automatically if the image is present (assuming the image is only rendered if the flag needs to be displayed).

Cue/app/tabs/library/CardListViewRow.js
```diff
@@ -21,24 +22,48 @@ const styles = {
...
+          style={{...styles.flagIcon, opacity: this.props.card.needs_review ? 1 : 0}}
+          source={CueIcons.indicatorFlag} />
       </View>
     )
+
+    // Row should not be touchable if flagCard is undefined.
+    if (typeof this.props.flagCard === 'undefined') return row
+
+    return (
+      <TouchableNativeFeedback
```
mattyu007: `TouchableNativeFeedback` isn't supported on iOS, you'll need to use something else for non-Android platforms. I recommend `TouchableHighlight` with `underlayColor={CueColors.veryLightGrey}`.

Cue/app/tabs/library/CardListView.js
mattyu007: Like with `CardListViewRow`, since this is a callback, I would call this `onFlagCard` instead.

ClaytonPassmore:
@mattyu007 updated!

Cue/app/tabs/library/DeckView.js
```diff
@@ -301,6 +307,20 @@ function actions(dispatch) {
...
+        cards: [
+          {
+            action: 'edit',
+            uuid: cardUuid,
+            needs_review: flag,
+          }
+        ]
+      }
+      /* TODO: Use flag card action */
+      return dispatch(editDeck(change))
```
Tyler-Brenneman: #153 will have `flagCard` action

Tyler-Brenneman (approved):
Looks good! Might want to wait for #153 to use the flag action

ClaytonPassmore:
Waiting for #153

Cue/app/tabs/library/CardListViewRow.js
```diff
@@ -1,44 +1,91 @@
...
+          {flag}
+        </View>
         <Text style={styles.backText}>
           {this.props.card.back}
         </Text>
       </View>
     )
+
+    // Row should not be touchable if onFlagCard is undefined.
+    if (typeof this.props.onFlagCard === 'undefined') return row
```
mattyu007: Nitpick: It's sufficient to just check `if (!this.props.onFlagCard)`.

mattyu007 (approved):
Looks good. Don't forget to squash before merging.

henryc132 (approved):
Since everyone seems to be approving

PR #153, opened 2017-03-23
Use flag endpoint

### LINKED ISSUE #113: Use the flag cards endpoint instead of the edit deck endpoint when flagging cards

### LINKED ISSUE #133: Error when syncing card edits to a new card
1. Create new card
2. edit card
3. sync changes
4. 400 server response

This happens because we create 2 card changes `{[action:"add", front:"fox", back:"20XX"},{action:"edit", needs_review:true}]` and keep them as 2 separate actions in `localChanges`.

To fix this we will need to create an uuid for cards when creating them offline, and combine all changes into 1 action for a card.

Similar to #103

Tyler-Brenneman:
Closes #113 
Closes #133 

Use flag endpoint if edit change action is `flag`

adds a new `flagCard` action

### DIFF
--- a/Cue/app/actions/library.js
+++ b/Cue/app/actions/library.js
@@ -114,7 +114,7 @@ async function syncDeck(change) : PromiseAction {
       serverDeck = await LibraryApi.editDeck(change)
     } else if (change.cards && change.cards.find(card => card.uuid && card.needs_review != undefined)) {
       let response = await LibraryApi.flagCard(change)
-      change = {...change, User_data_version: response.user_data_version}
+      change = {...change, user_data_version: response.user_data_version}
     } else {
       console.warn('Change not synced', change)
     }
--- a/Cue/app/actions/library.js
+++ b/Cue/app/actions/library.js
@@ -95,11 +95,11 @@ function syncLibrary(localChanges): ThunkAction {
 // ex: { uuid: string, action: 'edit', cards: [{action: 'edit', uuid: string, front: "fox", back: "20XX"}] } if 1 card in deck was changed
 async function syncDeck(change) : PromiseAction {
   let serverDeck = {}
-  if (change.action === "add"){
+  if (change.action === "add") {
     serverDeck = await LibraryApi.createDeck(change.name, change.tags);
     // check if any other changes need to be synced
     for (let key in change) {
-      if((key === 'cards' && change.cards.length) || (key === change[key] != serverDeck[key] && !key.match("^(?:cards|uuid|action)$"))) {
+      if((key === 'cards' && change.cards.length) || (change[key] != serverDeck[key] && !key.match("^(?:cards|uuid|action)$"))) {
         serverDeck = await LibraryApi.editDeck({
           ...change,
           uuid: serverDeck.uuid,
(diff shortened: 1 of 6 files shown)


Cue/app/actions/library.js
```diff
@@ -110,7 +110,15 @@ async function syncDeck(change) : PromiseAction {
       }
     }
   } else if (change.action === "edit") {
-    serverDeck = await LibraryApi.editDeck(change)
+    if (change.name || change.tags || change.public
+      || (change.cards && change.cards.find(card => card.front || card.back || card.poistion))) {
```
mattyu007: `position` and oof, this seems really fragile.

What happens if there are some cards which are edited and some which are flagged? Would it be better to introduce a new client-only action type `flag` to clearly separate the two actions?

Cue/app/actions/library.js
```diff
@@ -110,7 +110,15 @@ async function syncDeck(change) : PromiseAction {
...
+      serverDeck = await LibraryApi.editDeck(change)
+    } else if (change.cards && change.cards.find(card => card.uuid && card.needs_review != undefined)) {
+      let response = await LibraryApi.flagCard(change)
+      change = {...change, User_data_version: response.user_data_version}
```
mattyu007: Lowercase `u` on `user_data_version`.

Cue/app/actions/library.js
Tyler-Brenneman: Right now this will cause the server a 400 error.

2 ways to fix that.
1) we can either send 2 requests from the client, 1 for every all the card edits and 1 for all the flags.
2) server supports all action properties as optional, so that in 1 deck edit request we can edit 1 card in an action and just flag another

Cue/app/reducers/library.js
```diff
@@ -63,23 +63,52 @@ function library(state: State = initialState, action: Action): State {
...
+      }
+    }
+    let changeIndex = localChanges.findIndex(deck => deck.uuid == change.uuid)
+    if (localChanges[changeIndex]) {
+      let cards = localChanges[changeIndex].cards
+        ? change.cards
+          ? localChanges[changeIndex].cards.concat(change.cards)
+          : localChanges[changeIndex].cards
+        : change.cards
+
```
Tyler-Brenneman: this will be refactored when i do issue #133

Tyler-Brenneman:
Updated to use a `flagCard` action

ClaytonPassmore:
@lonebutterfish so if I flag a card and then delete it later before things get sync'd, is it possible that the card edit changes can occur before the flag changes, causing us to try and flag a deleted card? Conversely, what happens if we add a card and flag it before syncing? Does the flag request use the correct uuid?

Tyler-Brenneman:
@ClaytonPassmore  this pr doesn't fix that. My next pr will close #133 that will be fixed

mattyu007:
So right now if we just flag cards and then sync, we're fine. But if we flag the card and make any edits, then we have to call the edit deck API, which will give us a 400 since the flagged cards don't have `front`, `back`, or `position`.

If we naively make two separate API calls, I feel like we might end up with a sync conflict since both the edit and flag calls would be made with the same `parent_deck_version` and `parent_user_data_version`. How much work is it to change the edit deck API handler instead to support all properties being optional so we don't have to fudge the deck version on the client? @ClaytonPassmore or @henryc132?

ClaytonPassmore:
@mattyu007 ya it wouldn't be too much work, I can do it tomorrow.

Tyler-Brenneman:
updated PR to solve #133. Will combine all card changes into 1

mattyu007 (approved):

PR #154, opened 2017-03-23
Update DeckSharingOptions to use unshare when setting decks to private

### LINKED ISSUE #145: Update DeckSharingOptions to use the new unshare change key when converting decks to private
Right now we just set `share_code` to `null`, which doesn't actually work.

## Required updates
1. `DeckSharingOptions`
    * `setPublic` action should use the new `unshare` key
2. `api/Library.js`
    * `unshare` key needs to be added to the request body in `editDeck`

mattyu007:
Closes #145.

### DIFF
--- a/Cue/app/tabs/library/DeckView.js
+++ b/Cue/app/tabs/library/DeckView.js
@@ -254,6 +254,7 @@ class DeckView extends React.Component {
           onScroll={this._onScroll}>
           <DeckViewInfoHeader
             deck={this.state.deck}
+            key={this.state.deck.share_code}
             onLayout={this._onUpdateHeaderLayout} />
           <CardListView
             cards={this.state.deck.cards}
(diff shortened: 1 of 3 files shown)


Cue/app/tabs/library/DeckView.js
```diff
@@ -254,6 +254,7 @@ class DeckView extends React.Component {
           onScroll={this._onScroll}>
           <DeckViewInfoHeader
             deck={this.state.deck}
+            key={this.state.deck.share_code}
```
mattyu007, line 257: TBH I have no idea why `DeckViewInfoHeader` doesn't re-render when the deck is updated. It looks like it's receiving a stale copy of the deck, but I haven't been able to figure out why.

Tyler-Brenneman: does setting the key fix the problem? Tested it and it seems to update for me

mattyu007: Setting the key forces the header to unmount and remount every time the key changes. In this case, my issue was going from private to shared, where the header wouldn't update until you popped and re-pushed `DeckView` off and back on to the Navigator stack.

Tyler-Brenneman (approved):

PR #155, opened 2017-03-23
Fix UI bugs, connect discover to redux, and add loading notifications

### LINKED ISSUE #134: Show loading indicator when library is first being loaded
When a user first logs in, it loads the library async. Right now the library screen will show empty library view until the request finishes.

We could show a spinner while this loading takes place to show the user the data is still being fetched.

This can be done by checking if `librarydecks` is `null`. This means the decks hasn't been loaded yet.

### LINKED ISSUE #136: Card counter in DeckThumbnail is wrong until the deck is synced
If a deck is created locally, then the counter displays "undefined cards" until the deck is synced.

screen shot 2017-03-21 at 6 16 02 am

If the deck is updated locally, the counter does not update until the deck is synced.

`DeckThumbnail` now pulls the number of cards from the `num_cards` property, which is not created in the `createDeck` action and is not updated by the `editDeck` reducer. This was originally intended so that the number of cards could still be displayed even when we only had a deck's metadata (in Discover and Search), but we should consider only using `num_cards` as a fallback in case the `cards` property is not present.

henryc132:
Closes #136 
Closes #134 

Fixes:
- Fix card counter on thumbnail for newly created decks
- Fix background color on DeckPreview page.
- Connected DiscoverHome to the redux store.
- Added swipe to refresh on DiscoverHome
- Show spinner while Library or Discover page is loaded for first time

### DIFF
--- a/Cue/app/store/configureStore.js
+++ b/Cue/app/store/configureStore.js
@@ -20,8 +20,8 @@ var createCueStore = applyMiddleware(thunk, promise, logger)(createStore);
 
 function configureStore(onComplete: ?() => void) {
   const store = autoRehydrate()(createCueStore)(reducers);
-  persistStore(store, {storage: AsyncStorage, blacklist: ['tabs', 'discover']}, onComplete);
-  
+  persistStore(store, {storage: AsyncStorage, blacklist: ['tabs']}, onComplete);
+
   if (isDebuggingInChrome) {
     window.store = store;
   }
--- a/Cue/app/store/configureStore.js
+++ b/Cue/app/store/configureStore.js
@@ -20,7 +20,7 @@ var createCueStore = applyMiddleware(thunk, promise, logger)(createStore);
 
 function configureStore(onComplete: ?() => void) {
   const store = autoRehydrate()(createCueStore)(reducers);
-  persistStore(store, {storage: AsyncStorage, blacklist: ['tabs']}, onComplete);
+  persistStore(store, {storage: AsyncStorage, blacklist: ['tabs', 'discover']}, onComplete);
 
   if (isDebuggingInChrome) {
     window.store = store;
--- a/Cue/app/tabs/discover/DiscoverHome.js
+++ b/Cue/app/tabs/discover/DiscoverHome.js
@@ -1,7 +1,7 @@
 // @flow
 
 import React from 'react'
-import { View, Text, Image, ScrollView, ListView, RefreshControl, Navigator, Platform } from 'react-native'
+import { View, Text, Image, ScrollView, ListView, RefreshControl, Navigator, Platform, Alert } from 'react-native'
 import type { DeckMetadata } from '../../api/types';
 
 import { discoverDecks } from '../../actions'
@@ -60,11 +60,25 @@ class DiscoverHome extends React.Component {
     this.setState(this._getNewState(newProps))
   }
 
+  _fetchDiscoverDecks = () => {
+    this.props.onRefreshDiscover()
+    .catch(e => {
+      console.warn('Failed to fetch discover decks', e)
+      this.setState({
+        refreshing: false,
+      });
+      Alert.alert(
+        (Platform.OS === 'android' ? 'Failed to fetch Discovery decks' : 'Failed to Fetch Discovery Decks'),
+        'Check your Internet connection and try again.'
+      )
+    })
+  }
+
   _refresh = () => {
     this.setState({
       refreshing: true
     })
-    this.props.onRefreshDiscover()
+    this._fetchDiscoverDecks()
   }
 
   _getNewState = (props) => {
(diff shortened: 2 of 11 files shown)


Cue/app/tabs/library/LibraryHome.js
```diff
@@ -66,10 +66,12 @@ class LibraryHome extends React.Component {
...
   constructor(props: Props) {
     super(props)
-
+
+    let refreshing = this.props.decks ? false : true;
+
     this.state = {
       editing: false,
-      refreshing: false,
+      refreshing: refreshing,
```
mattyu007: You don't need to write `refreshing: refreshing`, you can just write `refreshing`.

Cue/app/store/configureStore.js
```diff
@@ -20,8 +20,8 @@ var createCueStore = applyMiddleware(thunk, promise, logger)(createStore);
 
 function configureStore(onComplete: ?() => void) {
   const store = autoRehydrate()(createCueStore)(reducers);
-  persistStore(store, {storage: AsyncStorage, blacklist: ['tabs', 'discover']}, onComplete);
-  
+  persistStore(store, {storage: AsyncStorage, blacklist: ['tabs']}, onComplete);
```
mattyu007: Do we want search results and the Discover page to persist across sessions?

Cue/app/tabs/discover/DiscoverHome.js
```diff
@@ -25,37 +28,61 @@ const styles = {
 
 type Props = {
   navigator: Navigator,
-  onPressMenu?: () => void
+  onPressMenu?: () => void,
+
+  newDecks: ?Array<DeckMetadata>,
+  topDecks: ?Array<DeckMetadata>,
+  onDiscover: () => any,
```
mattyu007: `onRefreshDiscover`?

Cue/app/tabs/discover/DiscoverHome.js
```diff
@@ -25,37 +28,61 @@ const styles = {
...
 }
 
 class DiscoverHome extends React.Component {
   props: Props
 
   state: {
-    dataSource: ListView.DataSource
+    dataSource: ListView.DataSource,
+    loading: boolean,
```
mattyu007: `loading` is not used?

mattyu007 (commented):
Looks good, just a few minor things.

It was so hard to verify that the loading spinner actually came up for the Library and Discover home pages since the loading time is so short, but indeed they do show up if I use a network link conditioner to simulate a 2G connection 😝

henryc132:
@mattyu007 werent you going to sleep? lol

Cue/app/tabs/discover/DiscoverHome.js
henryc132: Yea, forgot to remove it

Cue/app/store/configureStore.js
henryc132: Yea, it wouldn't make sense to show really old data

Cue/app/tabs/discover/DiscoverHome.js
```diff
@@ -25,37 +28,60 @@ const styles = {
...
+
+  componentWillReceiveProps(newProps) {
+    this.setState(this._getNewState(newProps))
+  }
+
+  _refresh = () => {
+    this.setState({
+      refreshing: true
+    })
+    this.props.onRefreshDiscover()
```
Tyler-Brenneman: Should add a catch here and display an alert or something that there is no internet connection

Tyler-Brenneman (commented):
Should handle user not having internet connection.

Same change should probably be made to search screen as well

Cue/app/tabs/library/LibraryHome.js
```diff
@@ -66,16 +66,18 @@ class LibraryHome extends React.Component {
 
   constructor(props: Props) {
     super(props)
-
+
+    let refreshing = this.props.decks ? false : true;
```
Tyler-Brenneman: setting `refreshing` here will cause it to stay `true` on login. Need a way to set it to `false` once the initial `loadLibrary` after login is done.

henryc132: weird, I swore there used to be a componentWillReceiveProps function in library home

Tyler-Brenneman: Oh, I removed it since a `syncDeck` action would trigger it and stop the refreshing icon before `loadLibrary` was finished

henryc132: But doesnt sync deck always happen before loadLibrary

Cue/app/tabs/library/LibraryHome.js
```diff
@@ -65,17 +65,26 @@ class LibraryHome extends React.Component {
...
-
-  _refresh = () => {
+    super(props)
+
+    let refreshing = false
+
+    if (!this.props.deck) {
+      refreshing = true
+      this.props.onLoadLibrary().then(response => {
+        this.setState({refreshing: false})
```
henryc132, line 75: Wonder what the probability of this being ran before we set state below is.

Tyler-Brenneman (approved):

PR #157, opened 2017-03-23
Set name on login

ClaytonPassmore:
- Name was previously not being set on Cue server
- Send Facebook name to Cue
- Cleaned up some of the login code
- Fix JSON response from server to match API spec
  - Now response contains `user_id` and `access_token` instead of `user-id` and `access-token`


Cue/app/actions/login.js
```diff
@@ -26,35 +29,32 @@ function logOut(): Action {
...
     name,
   };
 }
 
+
 function serverLogin(): ThunkAction {
   return (dispatch, getState) => {
-    const login = _serverLogin();
+    return _serverLogin().then(cueUser => {
+      dispatch(logIn(cueUser['user_id'], cueUser['access_token']));
```
ClaytonPassmore, line 44: If you want to test this against the server, you're going to have to change it to `user-id` and `access-token` until the server gets updated

Cue/app/actions/login.js
```diff
@@ -26,35 +29,32 @@ function logOut(): Action {
 }
 
 
-
-async function loadUsername(name): PromiseAction {
+async function loadUsername(name: string): PromiseAction {
```
mattyu007: Minor nitpick, but since this function isn't actually async, you can remove the `async` declaration and update the return type to just `Action`.

mattyu007 (approved):
Great work.

PR #158, opened 2017-03-23
Fix loadLibrary bugs

Tyler-Brenneman:
Fix 2 bugs

- don't `loadLibrary` on logout

- update refresh spinner only when `loadLibrary` is done

- edit button not showing up after syncing changes

### DIFF
--- a/Cue/app/tabs/library/LibraryHome.js
+++ b/Cue/app/tabs/library/LibraryHome.js
@@ -73,12 +73,6 @@ class LibraryHome extends React.Component {
     }
   }
 
-  componentWillReceiveProps(newProps: Props) {
-    this.setState({
-      refreshing: false
-    })
-  }
-
   _refresh = () => {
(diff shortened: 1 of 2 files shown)


Cue/app/tabs/library/LibraryHome.js
```diff
@@ -80,21 +80,20 @@ class LibraryHome extends React.Component {
   }
 
   _refresh = () => {
+    this.setState({refreshing: true})
     this.props.onSyncLibrary(this.props.localChanges).then(failedSyncs =>{
       if (failedSyncs && failedSyncs.length) {
         this.props.navigator.push({failedSyncs})
+        this.setState({refreshing: false})
```
mattyu007: I don't think it really matters, but I'd call `setState` before pushing the navigator route.

mattyu007 (approved):

PR #159, opened 2017-03-24
Fix empty public decks message

### LINKED ISSUE #123: "Add a card by tapping the + button" appears for empty decks in Preview Public Deck
We should not display the next step suggestion message for decks viewed from Preview Public Deck.

screen shot 2017-03-19 at 2 49 02 am

Tyler-Brenneman:
Closes #123

### DIFF
--- a/Cue/app/tabs/library/CardListView.js
+++ b/Cue/app/tabs/library/CardListView.js
@@ -13,6 +13,7 @@ import ListViewHairlineSeparator from '../../common/ListViewHairlineSeparator'
 type Props = {
   cards: Array<Card>,
   isFiltering: boolean,
+  accession: ?string,
 }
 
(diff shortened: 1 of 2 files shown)


Cue/app/tabs/library/CardListView.js
```diff
@@ -13,6 +13,7 @@ import ListViewHairlineSeparator from '../../common/ListViewHairlineSeparator'
 type Props = {
   cards: Array<Card>,
   isFiltering: boolean,
+  accession: ?string,
```
mattyu007: This actually needs to be `accession?: string` for Flow to type-check this correctly since the `accession` prop itself might be completely missing, rather than the prop occasionally having a `null` value.

mattyu007 (approved):
Other than the type thing, looks good.

PR #160, opened 2017-03-24
Alerts for deck name checking

Tyler-Brenneman:
Checks to make sure a new deck name is a valid length, and displays an alert if it isn't.

Let me know if you think the alert wording should be different.

### DIFF
--- a/Cue/app/tabs/library/LibraryHome.js
+++ b/Cue/app/tabs/library/LibraryHome.js
@@ -165,7 +165,9 @@ class LibraryHome extends React.Component {
     }
   }
 
+
   _onPressCreateDeck = () => {
+    const MAX_LENGTH = 255
     CuePrompt.prompt(
       Platform.OS === 'android' ? 'Create new deck' : 'Create New Deck',
       '',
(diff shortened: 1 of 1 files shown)


Cue/app/tabs/library/LibraryHome.js
```diff
@@ -111,7 +111,16 @@ class LibraryHome extends React.Component {
       [
         {text: 'Cancel', style: 'cancel'},
         {text: 'Create', onPress: (deckName) => {
-          if (deckName && deckName.length) {
+          if (!deckName || !deckName.length) {
+            Alert.alert(
+              Platform.OS === 'android' ? 'Enter a deck name' : 'Enter a Deck Name',
```
mattyu007: -----
**Deck Name Can’t Be Empty**

A good deck name is short but descriptive.

-----
? I'm not sure a second sentence is really necessary though. Need a second opinion.

Cue/app/tabs/library/LibraryHome.js
```diff
@@ -111,7 +111,16 @@ class LibraryHome extends React.Component {
...
+            )
+          } else if (deckName.length > 255) {
+            Alert.alert(
+              Platform.OS === 'android' ? 'Deck name too long' : 'Deck Name Too Long',
+              'Please enter a shorter deck name',
```
mattyu007: -----
**Deck Name Too Long**

Use a shorter name for your deck.

-----
? I don't feel too strongly about either, but I'd avoid using the word "please".

mattyu007 (commented):
Ugh, I wish we could just disable the Create button instead of needing to use more alerts.

Question: Should we re-display the alert if a validation error occurs?

Cue/app/tabs/library/LibraryHome.js
ClaytonPassmore: Yeah I feel like the title gets the message across

Cue/app/tabs/library/LibraryHome.js
ClaytonPassmore: pls

mattyu007 (approved):

PR #167, opened 2017-03-24
Make Deck Edit Options Dynamic

### LINKED ISSUE #163: Make deck edit actions dynamic
For any edit action, only require the part of the card being modified.

ClaytonPassmore:
- Made all parts of the edit action optional so that you only modify the part of the card that you need to modify
- Decks will now be created with empty tags rather than an empty string
- Added a migration for all decks that have an empty string as their tags delimited
- Removed `EDIT_CARD` procedure

Closes #163

### DIFF
--- a/Server/App/model/card.py
+++ b/Server/App/model/card.py
@@ -16,9 +16,23 @@ class Card(object):
         connector.call_procedure_transactionally(procedure_name, uuid, deck_id, front, back, position)
         return Card(uuid, connector)
 
-    def edit(self, deck_id, front, back, position):
-        self.connector.call_procedure_transactionally(
-            'EDIT_CARD', self.uuid, deck_id, front, back, position)
+    def edit(self, deck_id, front=None, back=None):
+        query = 'UPDATE Card SET {} WHERE deck_id=%s AND uuid=%s'
+        sub_list = []
+        arg_list = []
+        if front is not None:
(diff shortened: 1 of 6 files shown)


Server/App/model/card.py
```diff
@@ -16,9 +16,23 @@ def create(deck_id, front, back, position, reposition=True, connector=None):
...
+        if front is not None:
+            sub_list.append('front=%s')
+            arg_list.append(front)
+        if back is not None:
+            sub_list.append('back=%s')
+            arg_list.append(back)
+
+        if len(sub_list) > 0:
+            self.connector.query_transactionally(query.format(','.join(sub_list)),
+                                                 *(arg_list + [deck_id, self.uuid]))
```
henryc132, line 32: Having all queries in procedures would be nice, but ones with all these conditions would probably be annoying to write in sql

ClaytonPassmore: Yeah for sure, I kinda figured that given how it is dynamic it would just be easier to write in Python... maybe not a good long-term solution though.

henryc132 (approved):

Server/App/model/card.py
henryc132: Although now that I think about it, you could just select the card first, and then update the card to be the new value, provided param is not null, or set it back to the old value if its null. So its not too bad

PR #168, opened 2017-03-24
New Deck Play Options

### LINKED ISSUE #101: Add additional play modes to Play Deck
Play Deck needs two additional play modes:
1. Play the deck with the cards answer-side-up
2. Play through flagged cards only

ClaytonPassmore:
- Added new play option "With flagged cards only"
  - This option is automatically selected if the user was viewing the deck with the filter option turned on
- Added toggle-able modifiers (just one for now: "Answer side first")
- Made a new table row class for switches

Closes #101

Comments:
- No idea what this is going to look like on iOS.
- Style suggestions welcome, I just threw together what looked right to me
- Is the navigator stuff too messy?
- Thought modifiers would be nice since you might want to use answer-side-first with different play modes
  - Have to be careful about modifiers you make (i.e. make sure they don't conflict with play options / other modifiers)
- Do we need to be slicing our already sliced copy of the cards in `PlayDeckView`?

Modifier toggle off:
playsetup1

Modifier toggle on:
playsetup2

Picker in the middle:
playsetup3

### DIFF
--- a/Cue/app/common/CueColors.js
+++ b/Cue/app/common/CueColors.js
@@ -5,6 +5,7 @@
 module.exports = {
   primaryAccent: '#51A7F9',
 
+  primaryTintEvenLighter: '#BED1EA',
   primaryTintLighter: '#DEE8F5',
   primaryTint: '#0365C0',       // or Material Blue 600 '#1E88E5'?
   primaryTintDark: '#225095',   // or Material Blue 800 '#1565C0'?
--- a/Cue/app/common/CueColors.js
+++ b/Cue/app/common/CueColors.js
@@ -5,8 +5,8 @@
 module.exports = {
   primaryAccent: '#51A7F9',
 
-  primaryTintEvenLighter: '#BED1EA',
   primaryTintLighter: '#DEE8F5',
+  primaryTintSlightlyLighter: '#BED1EA',
   primaryTint: '#0365C0',       // or Material Blue 600 '#1E88E5'?
   primaryTintDark: '#225095',   // or Material Blue 800 '#1565C0'?
   primaryTintDarker: '#193C70', // or Material Blue 900 '#0D47A1'?
--- a/Cue/app/common/SwitchTableRow.js
+++ b/Cue/app/common/SwitchTableRow.js
@@ -1,0 +1,49 @@
+// @flow
+
+import React from 'react'
+import { Platform, Switch, Text } from 'react-native'
+
+import CueColors from './CueColors'
+
+import TableRow from './TableRow'
+
+
+const styles = {
+  container: {
+    flex: 1,
+    flexDirection: 'row',
+    alignItems: 'center',
+    justifyContent: 'space-between',
+    paddingVertical: Platform.OS === 'android' ? 16 : undefined,
+  },
+  text: {
+    color: CueColors.primaryText,
+    fontSize: Platform.OS === 'android' ? 16 : 17,
+  },
+}
+
+
+export default class SwitchTableRow extends React.Component {
+  props: {
+    text: string,
+    value: boolean,
+    onPress?: (value: boolean) => void,
+  }
+
+  render() {
+    return (
+      <TableRow style={styles.container} >
+        <Text style={styles.text}>
+          {this.props.text}
+        </Text>
+        <Switch
+          onTintColor={CueColors.primaryTint}
+          tintColor={CueColors.lightGrey}
+          thumbTintColor={(Platform.OS === 'android' && this.props.value) ?
+                            CueColors.primaryTint : CueColors.veryLightGrey}
+          value={this.props.value}
+          onValueChange={this.props.onPress} />
+      </TableRow>
+    )
+  }
+}
--- a/Cue/app/common/SwitchTableRow.js
+++ b/Cue/app/common/SwitchTableRow.js
@@ -32,15 +32,17 @@ export default class SwitchTableRow extends React.Component {
 
   render() {
     return (
-      <TableRow style={styles.container} >
+      <TableRow
+        style={styles.container}
+        onPress={() => this.props.onPress(!this.props.value)}
+        disabled={Platform.OS !== 'android'} >
         <Text style={styles.text}>
           {this.props.text}
         </Text>
         <Switch
-          onTintColor={CueColors.primaryTint}
-          tintColor={CueColors.lightGrey}
+          onTintColor={Platform.OS === 'android' ? CueColors.primaryTintEvenLighter : CueColors.primaryTint}
           thumbTintColor={(Platform.OS === 'android' && this.props.value) ?
-                            CueColors.primaryTint : CueColors.veryLightGrey}
+                            CueColors.primaryTint : undefined}
           value={this.props.value}
           onValueChange={this.props.onPress} />
       </TableRow>
--- a/Cue/app/common/SwitchTableRow.js
+++ b/Cue/app/common/SwitchTableRow.js
@@ -34,7 +34,7 @@ export default class SwitchTableRow extends React.Component {
     return (
       <TableRow
         style={styles.container}
-        onPress={() => this.props.onPress(!this.props.value)}
+        onPress={() => this.props.onPress && this.props.onPress(!this.props.value)}
         disabled={Platform.OS !== 'android'} >
         <Text style={styles.text}>
           {this.props.text}
--- a/Cue/app/common/SwitchTableRow.js
+++ b/Cue/app/common/SwitchTableRow.js
@@ -40,7 +40,7 @@ export default class SwitchTableRow extends React.Component {
           {this.props.text}
         </Text>
         <Switch
-          onTintColor={Platform.OS === 'android' ? CueColors.primaryTintEvenLighter : CueColors.primaryTint}
+          onTintColor={Platform.OS === 'android' ? CueColors.primaryTintSlightlyLighter : CueColors.primaryTint}
           thumbTintColor={(Platform.OS === 'android' && this.props.value) ?
                             CueColors.primaryTint : undefined}
           value={this.props.value}
--- a/Cue/app/tabs/library/play/PlayDeckSetupView.js
+++ b/Cue/app/tabs/library/play/PlayDeckSetupView.js
@@ -1,7 +1,7 @@
 // @flow
 
 import React from 'react'
-import { View, Text, ScrollView, Navigator, StyleSheet, Platform, Picker } from 'react-native'
+import { View, Text, ScrollView, Navigator, StyleSheet, Platform, Picker, Alert } from 'react-native'
 
 import type { Deck, Card } from '../../../api/types'
 
@@ -32,7 +32,7 @@ type Props = {
   flagFilter: boolean,
 }
 
-type PlaybackOption = 'sequential' | 'shuffled' | 'custom' | 'flagged'
+type PlaybackOption = 'sequential' | 'shuffled' | 'custom'
 
 export default class PlayDeckSetupView extends React.Component {
   props: Props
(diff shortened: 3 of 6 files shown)


Cue/app/common/SwitchTableRow.js
```diff
@@ -0,0 +1,49 @@
...
+export default class SwitchTableRow extends React.Component {
+  props: {
+    text: string,
+    value: boolean,
+    onPress?: (value: boolean) => void,
+  }
+
+  render() {
+    return (
+      <TableRow style={styles.container} >
```
mattyu007: Interaction nitpick: On Android, typically if there's a row with a switch, tapping anywhere on the row will flip the switch. It looks kinda funny if you also keep this interaction on iOS since the `Switch` on iOS doesn't animate value changes for some reason, but I think that's not too bad.

Cue/app/common/SwitchTableRow.js
```diff
@@ -0,0 +1,49 @@
...
+          {this.props.text}
+        </Text>
+        <Switch
+          onTintColor={CueColors.primaryTint}
+          tintColor={CueColors.lightGrey}
+          thumbTintColor={(Platform.OS === 'android' && this.props.value) ?
+                            CueColors.primaryTint : CueColors.veryLightGrey}
+          value={this.props.value}
+          onValueChange={this.props.onPress} />
+      </TableRow>
```
mattyu007: Style nitpicks:

* Leaving `tintColor` unspecified and using the platform defaults is fine, `lightGrey` gives the switch on iOS a really dark border
* Defining `thumbTintColor` for iOS makes the default shadow on the switch go away, which makes the switch kind of blend in instead of looking like a switch:
screen shot 2017-03-24 at 4 11 12 pm
screen shot 2017-03-24 at 4 13 19 pm


* Use a lighter tint like `'#BED1EA'` (which is `primaryTint` at 30% opacity) for the `onTintColor` for Android to keep the switch track distinct from the grabber:
screen shot 2017-03-24 at 4 08 38 pm

Cue/app/tabs/library/play/PlayDeckView.js
```diff
@@ -145,6 +153,24 @@ class PlayDeckView extends React.Component {
     StatusBar.setHidden(false)
   }
 
+  _flaggedOnly(array: Array<*>): Array<*> {
+    return array.slice(0).filter(item => item.needs_review)
```
mattyu007: A `slice` isn't necessary here since `filter` doesn't mutate the array

Cue/app/tabs/library/play/PlayDeckView.js
```diff
@@ -145,6 +153,24 @@ class PlayDeckView extends React.Component {
...
+
+    for (let i = 0; i < array.length; i++) {
+      ret.push({
+        ...array[i],
+        front: array[i].back,
+        back: array[i].front
+      })
+    }
+
+    return ret
```
mattyu007: You should be able to simplify this to `return array.map(deck => {...deck, front: deck.back, back: deck.front})`?

Cue/app/tabs/library/play/PlayDeckView.js
```diff
@@ -145,6 +153,24 @@ class PlayDeckView extends React.Component {
     StatusBar.setHidden(false)
   }
 
+  _flaggedOnly(array: Array<*>): Array<*> {
```
mattyu007: I would actually specify the generic parameter here and in `_answerFirst` as `Deck` since we're actually requiring that this is an array of `Deck` objects, rather than just any old array of objects.

Cue/app/tabs/library/play/PlayDeckSetupView.js
```diff
@@ -103,6 +111,19 @@ export default class PlayDeckSetupView extends React.Component {
     }
   }
 
+  _renderModifiers = () => {
+    return (
+      <View>
+        <TableHeader
+          text={'Modifiers:'} />
```
mattyu007: "Options" is a better word here, I think.

mattyu007 (commented):
Look great. A few comments, plus I guess we're still working out the interaction for this screen.

Cue/app/tabs/library/play/PlayDeckView.js
ClaytonPassmore: HA. yes.

Cue/app/common/SwitchTableRow.js
ClaytonPassmore: This was exactly the help I needed :stuck_out_tongue:

ClaytonPassmore:
Going to change flow a bit so DNM

Cue/app/common/CueColors.js
```diff
@@ -5,6 +5,7 @@
 module.exports = {
   primaryAccent: '#51A7F9',
 
+  primaryTintEvenLighter: '#BED1EA',
```
mattyu007: Actually this colour is darker than `primaryTintLighter` (which is `primaryTint` at 15% opacity and is the background colour for the user name in the drawer menu and also as the background colour for tags on iOS). 😬

ClaytonPassmore: LOL oops... Naming scheme is gonna have to get creative

mattyu007: `primaryTintLighter`
`primaryTintOnlyALittleBitLighter`
`primaryTint`

[Maybe we should have gone with numbers instead](https://material.io/guidelines/style/color.html#color-color-palette).

ClaytonPassmore:
Ok, so I updated the interactions a bit.. take a look at the screen caps

Left the shuffle and custom start mutex because having shuffle and custom start on doesn't make any sense:
opt1

Custom start:
opt2

If you have custom start selected but no cards to choose from (because flagged cards only is on and none of your cards are flagged), don't display the picker:
opt3

If you try to play a deck with flagged cards only when there are no flagged cards, you get this dialogue:
opt4

Cue/app/common/CueColors.js
```diff
@@ -6,6 +6,7 @@ module.exports = {
   primaryAccent: '#51A7F9',
 
   primaryTintLighter: '#DEE8F5',
+  primaryTintSlightlyLighter: '#BED1EA',
```
mattyu007, line 9: I'm actually dying 😂

ClaytonPassmore:
Also full disclaimer: Flow broke on my laptop so ... ya

Cue/app/tabs/library/play/PlayDeckSetupView.js
```diff
@@ -37,23 +39,88 @@ export default class PlayDeckSetupView extends React.Component {
...
-    this.state = {
+    let state = {
       playbackOption: 'sequential',
-      customStartIndex: 0
+      flaggedOnly: this.props.flagFilter,
+      answerFirst: false,
+      customStartIndex: 0,
+    }
+    state.filteredCards = this._generateDeckFilter(state)(this.props.deck.cards || [])
+    this.state = state
```
mattyu007: ~Why not just combine this into a single assignment? You can access `this.state` directly in `_generateDeckFilter` so you don't have to recreate the filter all the time.~

ohhhh ok it makes sense now. I'd add a comment or something to make this immediately obvious.

Cue/app/tabs/library/play/PlayDeckSetupView.js
```diff
@@ -133,6 +210,7 @@ export default class PlayDeckSetupView extends React.Component {
         <ScrollView style={styles.container}>
```
mattyu007: Can you add like, 40 pt of bottom padding to the `ScrollView`'s `contentContainerStyle` so that the bottom rows aren't exactly touching the bottom of the screen?

screen shot 2017-03-25 at 6 17 36 pm

Cue/app/tabs/library/play/PlayDeckSetupView.js
```diff
@@ -37,23 +39,88 @@ export default class PlayDeckSetupView extends React.Component {
...
+    newState.filteredCards = this._generateDeckFilter(newState)(this.props.deck.cards || [])
+
+    this.setState(newState)
+  }
+
+  _onPressPlay = () => {
+    if (this.state.filteredCards.length === 0) {
+      Alert.alert(
+        Platform.OS === 'android' ? 'No cards to play' : 'No Cards To Play',
+        'Try turning off the "Flagged cards only" option'
```
mattyu007: Curly quotes and a period at the end.

Cue/app/tabs/library/play/PlayDeckSetupView.js
```diff
@@ -37,23 +39,88 @@ export default class PlayDeckSetupView extends React.Component {
...
 
-    this.state = {
+    let state = {
       playbackOption: 'sequential',
-      customStartIndex: 0
+      flaggedOnly: this.props.flagFilter,
+      answerFirst: false,
+      customStartIndex: 0,
+    }
+    state.filteredCards = this._generateDeckFilter(state)(this.props.deck.cards || [])
```
mattyu007: Sort of dumb, but if you want this to type-check with Flow, you'll need to pre-create the `filteredCards` key in the object literal and assign it a dummy `[]` value before immediately overwriting it with the filtered cards.

Cue/app/tabs/library/play/PlayDeckSetupView.js
```diff
@@ -37,23 +39,88 @@ export default class PlayDeckSetupView extends React.Component {
...
+    if (this.state.filteredCards.length === 0) {
+      Alert.alert(
+        Platform.OS === 'android' ? 'No cards to play' : 'No Cards To Play',
+        'Try turning off the "Flagged cards only" option'
+      )
+      return
+    }
+    this.props.navigator.replace({
+      playDeck: this.props.deck,
+      cardFilter: this._generateDeckFilter(this.state),
```
mattyu007: This is kinda weird but I guess it makes sense since we need access to other properties from the deck in `PlayDeckView`.

Cue/app/common/SwitchTableRow.js
```diff
@@ -0,0 +1,51 @@
...
+    text: string,
+    value: boolean,
+    onPress?: (value: boolean) => void,
+  }
+
+  render() {
+    return (
+      <TableRow
+        style={styles.container}
+        onPress={() => this.props.onPress(!this.props.value)}
```
mattyu007: Since `onPress` is optional, Flow will require you do do a check to make sure it's not undefined before calling it, i.e., `() => this.props.onPress && this.props.onPress(!this.props.value)`.

mattyu007 (commented):
New flow looks fine. It's still a lot of switches, but we did commit to a lot of features in play mode, so...

Cue/app/tabs/library/play/PlayDeckSetupView.js
ClaytonPassmore: Yeah I don't know man :man_shrugging:

Cue/app/tabs/library/play/PlayDeckSetupView.js
ClaytonPassmore: Is that just by doing `paddingBottom: 40`? Because `paddingBottom: '40pt'` does not work

mattyu007: No, there's a separate `contentContainerStyle` prop you have to pass in to ScrollView. 

But yeah, something like `contentContainerStyle={{paddingBottom: 40}}`

ClaytonPassmore: ohhhhhhh ok will update

mattyu007 (approved):
Great work. Don't forget to squash and rebase.

PR #172, opened 2017-03-24
Fix deck overwrite

### LINKED ISSUE #171: Server returns 400 on deck overwrite
- Issue introduced with #163 
- Need to change card edit methods in overwrite handler

ClaytonPassmore:
- Same change as in #167

Closes #171

### DIFF
--- a/Server/App/api/deck.py
+++ b/Server/App/api/deck.py
@@ -182,13 +182,8 @@ class DeckUUIDHandler(BaseHandler):
                     if card_obj is None:
                         raise ValueError  # not allowed to edit this card
 
-                    # Make edits.
-                    if card.get('front') is not None or card.get('back') is not None:
-                        card_obj.edit(deck.uuid, card.get('front'), card.get('back'))
-
-                    # Move card.
-                    if card.get('position') is not None:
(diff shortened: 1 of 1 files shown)


Server/App/api/deck.py
```diff
@@ -181,7 +181,14 @@ def post(self, name, public, tags, cards, device, user_id, connector, uuid):
...
                         raise ValueError  # not allowed to edit this card
-                    card_obj.edit(deck.uuid, card['front'], card['back'], card['position'])
+
+                    # Make edits.
+                    if card.get('front') is not None or card.get('back') is not None:
+                        card_obj.edit(deck.uuid, card.get('front'), card.get('back'))
+
+                    # Move card.
+                    if card.get('position') is not None:
+                        card_obj.move(deck.uuid, card['position'])
```
henryc132: I sort of feel like that "front" "back" "position" should be expected? Since this endpoint is supposed to be for overwriting after all

ClaytonPassmore: Yeah, you're right! Will update

henryc132 (approved):

PR #162, opened 2017-03-24
Look up deck via share code

ClaytonPassmore:
- Added flow mentioned in spec
- Needed to add optional `shareCode` param to several functions
  - Without the share code, you aren't allowed to fetch the deck if it's just shared and not public

Take a look at the screenshots

Need some iOS QA pls.

New flow when hitting the '+' or 'Add' button:
shareadd1

Prompt to enter share code:
shareadd2

Deck preview:
sharedadd3

What happens if you enter a bad share code:
shareadd4

Closes #144

PR #174, opened 2017-03-24
Add ability to rate and copy decks

### LINKED ISSUE #126: Add ability to rate public decks

### LINKED ISSUE #170: Add option to copy deck on deckview page

henryc132:
Add overflow icon in android to rate+copy decks, and an additional button on IOS to rate.

Currently, no way to see how you previously rated a deck. Open to suggestions as to how to indicate it in UI.

copy

rate

Closes #170 
Closes #126

### DIFF
--- a/Cue/app/actions/library.js
+++ b/Cue/app/actions/library.js
@@ -140,7 +140,6 @@ async function syncDeck(change) : PromiseAction {
     change = {...change, user_data_version: response.user_data_version}
   } else if (change.action === "rate") {
     await LibraryApi.rateDeck(change.uuid, change.user_rating)
-    change = {...change}
   }
   return {
     type: 'DECK_SYNCED',
--- a/Cue/app/actions/library.js
+++ b/Cue/app/actions/library.js
@@ -64,6 +64,17 @@ function flagCard(deckUuid: string, cardUuid: string, needs_review: Boolean): Ac
   }
   return {
     type: 'CARD_FLAGGED',
+    change,
+  }
+}
+
+function rateDeck(uuid: string, user_rating: number): Action {
+  let change = {
+    uuid,
+    user_rating
+  }
+  return {
+    type: 'DECK_RATED',
     change
   }
 }
@@ -127,6 +138,9 @@ async function syncDeck(change) : PromiseAction {
   } else if (change.action === "flag") {
     let response = await LibraryApi.flagCard(change)
     change = {...change, user_data_version: response.user_data_version}
+  } else if (change.action === "rate") {
+    await LibraryApi.rateDeck(change.uuid, change.user_rating)
+    change = {...change}
   }
   return {
     type: 'DECK_SYNCED',
@@ -179,5 +193,5 @@ function copyDeck(localDeck: Deck) : Action {
 
 module.exports = {
   loadLibrary, createDeck, deleteDeck, editDeck, recordShareCode, resolveConflict,
-  syncDeck, syncLibrary, addLibrary, clearInaccessibleDecks, copyDeck, flagCard
+  syncDeck, syncLibrary, addLibrary, clearInaccessibleDecks, copyDeck, flagCard, rateDeck
 };
--- a/Cue/app/actions/types.js
+++ b/Cue/app/actions/types.js
@@ -20,7 +20,7 @@ export type Action =
 	| { type: 'DECK_ALREADY_IN_LIBRARY'}
 	| { type: 'DECK_ADDED_TO_LIBRARY', addedDeck: Deck}
 	| { type: 'CLEAR_INACCESSIBLE_DECKS'}
-	| { type: 'CARD_RATED', change: {uuid: string, userRating: number}}
+	| { type: 'DECK_RATED', change: {uuid: string, userRating: number}}
 	;
 
 export type Dispatch = (action: Action | ThunkAction | PromiseAction | Array<Action>) => any;
--- a/Cue/app/actions/types.js
+++ b/Cue/app/actions/types.js
@@ -20,7 +20,8 @@ export type Action =
 	| { type: 'DECK_ALREADY_IN_LIBRARY'}
 	| { type: 'DECK_ADDED_TO_LIBRARY', addedDeck: Deck}
 	| { type: 'CLEAR_INACCESSIBLE_DECKS'}
-;
+	| { type: 'CARD_RATED', change: {uuid: string, userRating: number}}
+	;
 
 export type Dispatch = (action: Action | ThunkAction | PromiseAction | Array<Action>) => any;
 export type GetState = () => Object;
--- a/Cue/app/tabs/library/DeckView.js
+++ b/Cue/app/tabs/library/DeckView.js
@@ -6,7 +6,7 @@ import React from 'react'
 import { View, Text, Image, ScrollView, ListView, Navigator, Platform, Alert } from 'react-native'
 
 import { connect } from 'react-redux'
-import { editDeck, flagCard } from '../../actions'
+import { editDeck, flagCard, copyDeck, rateDeck } from '../../actions'
 
 import type { Card, Deck } from '../../api/types'
 
@@ -55,6 +55,8 @@ type Props = {
   addCard: (deckUuid: string, front: string, back: string, position: number) => any,
   editCard: (deckUuid: string, cardUuid: string, front: string, back: string) => any,
   flagCard: (deckUuid: string, cardUuid: string, flag: boolean) => any,
+  rateDeck: (deckUuid: string, rating: number) => any,
+  copyDeck: (deck: Deck) => any
 }
 
 class DeckView extends React.Component {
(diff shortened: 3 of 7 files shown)


Cue/app/tabs/library/DeckView.js
```diff
@@ -92,6 +94,34 @@ class DeckView extends React.Component {
     this.props.flagCard(this.props.deckUuid, cardUuid, flag)
   }
 
+  _rateDeck = () => {
+    Alert.alert(
+      (Platform.OS === 'android'
+        ? 'Rate this deck'
+        : 'Rate This Deck'),
+      'Would you recommend this deck to others?',
```
henryc132, line 108: Phrasing sounds awkward, open to suggestion on rewording?

Cue/app/actions/library.js
```diff
@@ -127,6 +138,9 @@ async function syncDeck(change) : PromiseAction {
   } else if (change.action === "flag") {
     let response = await LibraryApi.flagCard(change)
     change = {...change, user_data_version: response.user_data_version}
+  } else if (change.action === "rate") {
+    await LibraryApi.rateDeck(change.uuid, change.user_rating)
+    change = {...change}
```
mattyu007: What does this do?

Cue/app/actions/types.js
```diff
@@ -20,7 +20,8 @@ export type Action =
 	| { type: 'DECK_ALREADY_IN_LIBRARY'}
 	| { type: 'DECK_ADDED_TO_LIBRARY', addedDeck: Deck}
 	| { type: 'CLEAR_INACCESSIBLE_DECKS'}
-;
+	| { type: 'CARD_RATED', change: {uuid: string, userRating: number}}
```
mattyu007: Should this be `DECK_RATED`?

Cue/app/tabs/library/DeckView.js
```diff
@@ -202,12 +237,22 @@ class DeckView extends React.Component {
         ]
       } else {
         return [
-          copyItem
+          copyItem,
+          rateItem
```
mattyu007, line 248: `rateItem` is missing from the argument deconstruction.

Cue/app/tabs/library/DeckView.js
```diff
@@ -92,6 +94,34 @@ class DeckView extends React.Component {
...
+  }
+
+  _copyDeck = () => {
+    Alert.alert(
+      (Platform.OS === 'android'
+        ? 'Are you sure you want to copy this deck?'
+        : 'Are You Sure You Want to Copy This Deck?'),
+      'Copying will create a private version of this deck in your library',
+      [
+        {text: 'Cancel', style: 'destructive'},
```
mattyu007: Use `style: 'cancel'` instead. `style: 'destructive'` is only used for actions which are dangerous / can result in loss of the user's data.

Cue/app/tabs/library/DeckView.js
```diff
@@ -92,6 +94,34 @@ class DeckView extends React.Component {
...
+        {text: 'Recommend',  onPress: () => this.props.rateDeck(this.state.deck.uuid, 1)},
+      ],
+
+    )
+  }
+
+  _copyDeck = () => {
+    Alert.alert(
+      (Platform.OS === 'android'
+        ? 'Are you sure you want to copy this deck?'
```
mattyu007: "Copy this deck?" or "Make a private copy of this deck?" is probably fine.

Cue/app/tabs/library/DeckView.js
```diff
@@ -92,6 +94,34 @@ class DeckView extends React.Component {
...
+        : 'Are You Sure You Want to Copy This Deck?'),
+      'Copying will create a private version of this deck in your library',
```
mattyu007: Maybe add a note that this copy won't receive any updates made to the deck by the original owner.

Cue/app/tabs/library/DeckView.js
```diff
@@ -92,6 +94,34 @@ class DeckView extends React.Component {
...
+  _rateDeck = () => {
+    Alert.alert(
+      (Platform.OS === 'android'
+        ? 'Rate this deck'
+        : 'Rate This Deck'),
+      'Would you recommend this deck to others?',
+      [
+        {text: 'Cancel', style: 'destructive'},
+        {text: 'Would not Recommend', onPress: () => this.props.rateDeck(this.state.deck.uuid, -1)},
+        {text: 'Recommend',  onPress: () => this.props.rateDeck(this.state.deck.uuid, 1)},
```
mattyu007: Same issue here as #162 — https://github.com/ClaytonPassmore/ProjectOrange/pull/162/files#diff-554e51b603c97565b81036119f86579fR135. You'll need to rearrange the buttons based on platform, though the Cancel button should stick to the bottom automatically on iOS if you specify `style: 'cancel'`.

Cue/app/tabs/library/DeckView.js
mattyu007: Yeah I'm not sure... **Recommend** and **Would Not Recommend** aren't too bad. You could also try **Yes, Would Recommend** and **No, Would Not Recommend**, though those might be too long.

mattyu007 (commented):
Looks good, just a few things.

Cue/app/actions/types.js
mattyu007: Also, why this extra level of indirection with `change`?

henryc132: Do you mean why I chose to use local changed? Wanted to allow for rating while offline

Cue/app/actions/library.js
henryc132: actually yea, I dont need that

henryc132:
updated

Cue/app/actions/types.js
mattyu007: No, I was just wondering why you pass in a `change` object instead of just passing in `uuid` and `userRating` separately, but I guess this makes sense.

This should be `DECK_RATED`, instead of `CARD_RATED`, right?

Cue/app/tabs/library/DeckView.js
```diff
@@ -92,6 +94,38 @@ class DeckView extends React.Component {
...
+  _copyDeck = () => {
+    Alert.alert(
+      (Platform.OS === 'android'
+        ? 'Copy this deck?'
+        : 'Copy This Deck?'),
+      'Copying will create a private version of this deck in your library.'
+        + ' The copied deck will no longer receive updates from the original owner.',
+      [
+        {text: 'Cancel', style: 'cancel'},
+        {text: 'Copy', onPress: () => this.props.copyDeck(this.state.deck)}
```
mattyu007: We should probably replace the top of the navigator stack with the new deck.

mattyu007 (commented):
Other than the two things, looks good. It's kind of weird that there's no feedback after rating a deck, but that's something we can re-consider later.

Cue/app/tabs/library/DeckView.js
mattyu007: Also: *Not* is capitalized in Title Case.

Cue/app/tabs/library/DeckView.js
henryc132: Not exactly sure how I would get the new deck back, to push it onto the navigator stack. Unless I create some redux object whose only purpose is to show the last copied deck?

mattyu007: `dispatch` returns the action returned by the action creator, so you should be able to extract the deck UUID from that.

henryc132:
fixed action type name, and pushed new deck onto navigator stack

Cue/app/tabs/library/DeckView.js
```diff
@@ -92,6 +94,41 @@ class DeckView extends React.Component {
...
+      (Platform.OS === 'android'
+        ? 'Copy this deck?'
+        : 'Copy This Deck?'),
+      'Copying will create a private version of this deck in your library.'
+        + ' The copied deck will no longer receive updates from the original owner.',
+      [
+        {text: 'Cancel', style: 'cancel'},
+        {text: 'Copy', onPress: () => {
+          let action = this.props.copyDeck(this.state.deck)
+          this.props.navigator.push({deck: action.deck})
```
mattyu007: I would do a `replace` here actually so that tapping the back button after copying the deck returns the user to the Library screen rather than to the shared / public deck.

Cue/app/tabs/library/DeckView.js
```diff
@@ -92,6 +94,41 @@ class DeckView extends React.Component {
     this.props.flagCard(this.props.deckUuid, cardUuid, flag)
   }
 
+  _rateDeck = () => {
+    let buttons = [
+      {text: 'Recommend',  onPress: () => this.props.rateDeck(this.state.deck.uuid, 1)},
+      {text: 'Would not Recommend', onPress: () => this.props.rateDeck(this.state.deck.uuid, -1)},
```
mattyu007: **Not** needs to be capitalized.

mattyu007 (approved):
Great work. Please fix the two nitpicks and then squash / rebase before merging.

PR #177, opened 2017-03-25
Fix header colour in launch screen on iOS

mattyu007:
Because of how colour management works on iOS vs in Interface Builder <sub>ಠ__ಠ</sub> the first time I set the colour of the fake header on the launch screen, it didn't actually match the colour of the real header... Should be fixed now.

Before:
simulator screen shot mar 25 2017 12 27 29 am

After:
simulator screen shot mar 25 2017 12 28 01 am

Actual colour for reference:
simulator screen shot mar 25 2017 12 27 06 am


henryc132:
We have fake headers?

mattyu007:
This is for the launch screen on iOS, which is the screen which is initially displayed by the system while waiting for the app to launch. If you kill the app on your phone and launch it, the blank screen with the strip of blue up top which only shows up for a second or so is the launch screen.

That blue strip is the "fake header".

henryc132 (approved):
The black text makes the new header look darker than the actual color

mattyu007:
Yeah, the actual colours should be the same this time though.

PR #178, opened 2017-03-25
Add Deck View Edit Mode

### LINKED ISSUE #64: Add Deck View edit mode
In Deck View edit mode, users can delete, rearrange, and enter the Edit Deck flow (#60).

UX Reference:
cue ux mocks 010
cue ux mocks 011

### LINKED ISSUE #103: Reorder local cards on position change
When a card position change happens, the other cards in the deck might need its position updated.

`_mergeCardChanges` in library reducer has a todo for this.

### LINKED ISSUE #119: Unify messaging for Shared / Public status of non-owned decks
On the Library home page, the client sorts decks in the list using the `accession` property, whereas the Deck View header uses the `private` and `share_code` properties of the deck to determine what status to display.

We should unify these two to present a coherent message to the user.

While we're in the file, we should also disconnect the component from the Redux store and rely on the `accession` property to determine if the deck is owned by the current user.

### LINKED ISSUE #173: Can't delete multiple cards at a time

mattyu007:
It's a big one! Closes #64, closes #103, closes #119, and closes #173.

## Summary
This pull request:

1. Adds Edit Mode to Deck View, supporting editing, rearranging, and deleting cards; and editing deck names.
2. Updates the `library` reducer to properly handle rearranging cards.
3. Disconnects `DeckViewInfoHeader` from the Redux store by using `accession` to determine if a deck is owned by the current user.

## Implementation Notes
* `DeckView` no longer has the collapsing header, due to undesirable interactions with `react-native-sortable-listview`. Maybe we can pursue this again in the future.
     * As a side effect, this means that the `EmptyView`s in `DeckView` are now centred and laid out properly again, so it's not a complete loss. 😜
* `LayoutAnimation` crashes the app on Android, so the transition to and from Edit mode isn't animated. 😭 
* Since the server performs card moves in order, on the client, we only need to deal with cases where a card is moved multiple times without a sync happening in between. 

### New Packages
* `react-native-sortable-listview` is a replacement for `ListView` which supports dragging and dropping to rearrange items.

## Known Issues
* Entering Edit Mode doesn't actually disable the swipe-to-pop gesture on iOS or hook into the back button on Android, so it's possible to pop `DeckView` off the stack without hitting the done button. Changes are saved anyway, but it's just a weird UX.

## Screenshots
Tapping the Edit button in the toolbar switches Deck View to Edit Mode. The transition is animated on iOS, but the animation crashes the app on Android, so no animations there. All toolbar items are removed, and the flag filter is also unset if previously set.
ezgif-1-8ec04cb8d5

Up close. Tapping in the text area in the header allows the deck to be renamed. Tapping the delete icon on the left deletes the item immediately. (There is no iOS-style second stage confirmation.) Each Text item is limited to two lines to keep the individual rows relatively short.

**Question: Should we hide the flag in edit mode?**
screen shot 2017-03-25 at 3 30 32 am

Tapping in the middle of the row opens that card in the Edit Card flow.
screen shot 2017-03-25 at 3 32 47 am

Touch and hold the grabber on the right to lift the row and move it around.
ezgif-3-c75b5ab7fa


ClaytonPassmore:
@mattyu007 just before I dive in, a couple comments:
- Should we make the card backgrounds white so that you can't see through them when you're dragging them around?
- Should we be able to "cancel" an edit? (I'm guessing that's a huge amount of work because you would have to buffer the changes before sending them off)

mattyu007:
@ClaytonPassmore 

* re: card backgrounds
    * iOS makes row backgrounds transparent while dragging, but they also put a pretty obvious shadow around the row to make it super obvious. (Wasn't able to get that to work for some reason.) I can give the rows like, 35% opacity white background and scale them up a little more to keep them distinct.
* re: cancelling edits
    * It's possible, but your intuition is correct in that it would be a lot of work.

Cue/app/common/CueColors.js
```diff
@@ -10,6 +10,8 @@ module.exports = {
   primaryTintDark: '#225095',   // or Material Blue 800 '#1565C0'?
   primaryTintDarker: '#193C70', // or Material Blue 900 '#0D47A1'?
 
+  dangerTint: '#FF3B30',
```
ClaytonPassmore, line 14: dangerzone

ClaytonPassmore:
@mattyu007 ok ya if you could up the opacity a touch I think that would be good!

Don't worry about cancelling edits, ain't nobody got time for that

mattyu007:
Alright y'all I have two choices

1. Elevation (+ opaque rows)
ezgif-3-79f17dc4fd

2. Opacity
ezgif-3-78f5a0ca54

Cue/app/reducers/library.js
```diff
@@ -162,46 +162,129 @@ function library(state: State = initialState, action: Action): State {
...
+        console.warn('Tried to delete non-existant card ' + change.uuid + '?')
+      }
+
+    } else if (change.action === 'edit') {
+      if (index !== -1) {
+        if (typeof change.position !== 'undefined' && change.position !== index) {
+          // The change has a move; do the move
+          positionUpdateNeeded = true
+          let card = mergedCards.splice(index, 1)[0]
+          card = {...card, ...change, action: undefined}
```
ClaytonPassmore, line 218: Why is the action undefined here?

ClaytonPassmore: ooo never mind I get it

Cue/app/reducers/library.js
```diff
@@ -162,46 +162,129 @@ function library(state: State = initialState, action: Action): State {
...
+          let oldPosition = index
+          let newPosition = change.position
+
+          console.debug('Processing move from ' + oldPosition + ' to ' + newPosition)
+          // Forward move from X to Y; subtract 1 from all pending changes
+          // with positions in [X, Y)
+          if (newPosition > oldPosition) {
+            mergedCards.forEach((card: Card) => {
+              if (typeof card.position !== 'undefined'
+                  && card.position >= oldPosition
```
ClaytonPassmore: Technically this only needs to be `>`, right? I mean its fine because we spliced out the card from there already, just checking my sanity

Cue/app/reducers/library.js
```diff
@@ -162,46 +162,129 @@ function library(state: State = initialState, action: Action): State {
...
+                  && card.position < newPosition) {
```
ClaytonPassmore: I think this needs to be `<=`

Cue/app/reducers/library.js
```diff
@@ -162,46 +162,129 @@ function library(state: State = initialState, action: Action): State {
...
+                console.debug('Foward: Card at ' + card.position + ' will be adjusted -1')
```
ClaytonPassmore: What is "Foward"?

ClaytonPassmore: Oh... "Forward"?

Cue/app/tabs/library/CardListView.js
```diff
@@ -10,50 +11,107 @@ import EmptyView from '../../common/EmptyView'
...
+      order,
+    }
+  }
+
+  _rowHasChanged = (r1: AnnotatedCard, r2: AnnotatedCard) => {
+    return r1._editing !== r2._editing
+      || r1.uuid !== r2.uuid
+      || r1.front !== r2.front
+      || r1.back !== r2.back
+      || r1.needs_review !== r2.needs_review
```
henryc132, line 97: I think r1 != r2 is enough here? We never edit a card without creating a completely new object

Cue/app/tabs/library/CardListView.js
```diff
@@ -10,50 +11,107 @@ import EmptyView from '../../common/EmptyView'
 import CardListViewRow from './CardListViewRow'
 import ListViewHairlineSeparator from '../../common/ListViewHairlineSeparator'
 
+const styles = {
+  liftedRowStyle: {
+    opacity: 1,
+    transform: [{scale: 1.02}],
+    backgroundColor: 'white',
+    overflow: 'visible',
```
ClaytonPassmore, line 19: What's the reason for this property? Is this for shadow or something? Because the content doesn't overflow out of the container, does it?

Cue/app/tabs/library/CardListView.js
```diff
@@ -10,50 +11,107 @@ import EmptyView from '../../common/EmptyView'
...
+  onDeleteCard?: (card: Card) => void,
+  onEditCard?: (card: Card) => void,
   onFlagCard?: (cardUuid: string, flag: boolean) => any,
+  onMoveCard?: (cardUuid: string, from: number, to: number) => any,
 }
 
+// Store the most recent edit mode so that when the edit mode changes,
+// we can force the entire list to re-render without unmounting and remounting
+// which would cause us to lose our scroll position.
+type AnnotatedCard = Card & {_editing: boolean}
```
ClaytonPassmore, line 49: Had to look this one up :laughing:

Cue/app/tabs/library/CardListView.js
mattyu007: Unfortunately `!=` and `!==` don't do deep object equality, so they'll always return false if the two objects are different, even if they represent the same thing.

e.g., `{a: true} == {a: true}` returns false

Cue/app/reducers/library.js
mattyu007: Yeah we want to remove the action from the change so that it doesn't appear in the resulting merged card.

Cue/app/reducers/library.js
mattyu007: It has to be `>=` since we spliced the object out of the array, so any card which moved up as a result of the change needs to be adjusted.

henryc132: Well, splicing the object out won't change the position properties of the cards, only the index

Cue/app/reducers/library.js
mattyu007: I don't think so, since if an object is at index `Y`, then since the new change is pushed onto the back of the array, when we process the changes, the object at index `Y` will just be bumped out of the way, which is what we expect.

Cue/app/reducers/library.js
mattyu007: Oops.

Cue/app/tabs/library/CardListView.js
mattyu007: Yeah, it's to stop the shadows (which extend outside the bounds of the view) from being clipped.

Cue/app/tabs/library/CardListView.js
mattyu007: I only know this because we use it to extend the `CardMetadata` type 😆

Cue/app/tabs/library/CardListView.js
henryc132: From what I understand of redux, if we didn't edit a card, the same object should be returned right? So the != would work,

Cue/app/tabs/library/CardListViewRow.js
```diff
@@ -53,39 +92,92 @@ export default class CardListViewRow extends React.Component {
...
     }
 
-    let row = (
-      <View style={styles.container}>
+    let deleteContainer
+    let handleContainer
+    if (this.props.editing) {
+      let deleteIcon = (
+        <View style={styles.deleteContainer}>
+          <View>
```
ClaytonPassmore: What's the reason for the second `View`?

Cue/app/reducers/library.js
mattyu007: OK I think there's another bug here, will push a commit after @ClaytonPassmore and you finish up the first pass.

Cue/app/tabs/library/DeckView.js
```diff
@@ -253,23 +285,25 @@ class DeckView extends React.Component {
         <CueHeader
           leftItem={leftItem}
           rightItems={rightItems}
-          key={this.state.isFiltering.toString() + this.state.scrollPosition.toString()}
-          containerStyles={{elevation: this.state.scrollPosition > this.state.headerHeight ? 4 : 0}} />
-        <ScrollView
-          style={{flex: 1}}
-          onScroll={this._onScroll}>
+          key={this.state.filtering.toString()} />
```
ClaytonPassmore, line 343: What is this prop doing?

ClaytonPassmore (commented):
@mattyu007 first pass looks good! I have some concerns about the card position updating but that's really the only thing. I'll check out what's going on with the server and then maybe we can isolate the client and see what (if any) changes need to be made.

Cue/app/tabs/library/CardListViewRow.js
mattyu007: Umm you know, sometimes you just want an extra `View`, you know?

(It was probably there for layout, but the layout works fine without it, so...)

Cue/app/tabs/library/DeckView.js
mattyu007: This sets the key on the header to work around the problem where some items would disappear from the header after the Navigator stack pops.

Cue/app/tabs/library/CardListView.js
mattyu007: These aren't the objects returned by the Redux store though, they're annotated with an extra property in `_getDataAndOrder`.

mattyu007:
OK I took another stab at the move reconciliation "algorithm" by working it out on paper.

For moves from `x` to `y` where `x < y`, we have to adjust all pending changes with positions in the range `(x, y]` by -1.

For moves from `x` to `y` where `x > y`, we don't have to do anything except make sure that the move is performed last, since all the cards will be bumped automatically.

ClaytonPassmore:
@mattyu007 

Definitely agree with the `x < y` scenario!

For `x > y`, doing it last will work, but I think that you could also adjust all cards in the range `[x, y)` by +1 (At least I think that's how we're doing it [server side](https://github.com/ClaytonPassmore/ProjectOrange/blob/master/Server/SQL/procedures.sql#L96)).

mattyu007:
@ClaytonPassmore That is what happens when we update the local library in `_applyCardChangesDecks`, but for `localChanges`, if we also adjusted the positions in the `x > y` case, I think we would end up creating gaps?

ClaytonPassmore:
@mattyu007 

Sorry ya, I didn't fully grasp the scenario at first.

When you say `x` moves to `y`, do you mean that `x` is the original position of the card or is `x` the position given the existing change in position? And I assume `y` is the final position that we want the card to be in...

mattyu007:
@ClaytonPassmore `y` is the destination position, and `x` is the last recorded destination

ClaytonPassmore:
@mattyu007 ok so after writing out some examples,  I found a scenario where the method that I suggested for `x > y` doesn't work. The method that you suggested seemed to work every time.

... So by engineering induction I think your method is correct :1st_place_medal:

ClaytonPassmore:
@mattyu007 I was never able to find a scenario where we had to adjust the positions for other cards (i.e. the `x < y` scenario). Are you sure we have to do -1 at all?

mattyu007:
@ClaytonPassmore We do, but I think my handling for the `x < y` scenario is *still* wrong 😜

I think the correct solution is to adjust cards in `(x, +inf)` by -1. Consider the following:
```js
localChanges = [
  A -> 0,
  B -> 1,
  C -> 2,
  D -> 3,
]
```
Now suppose card `A` is moved from position 0 to position 1 (so we have `[B, A, C, D]`). Since we indiscriminately move all new changes to the end, then this is the series of changes which would correctly generate the new list:
```js
localChanges = [
  B -> 0, // -1 applied to 1 since 1 \in (0, +inf)
  C -> 1, // -1 applied to 2 since 2 \in (0, +inf)
  D -> 2, // -1 applied to 3 since 3 \in (0, +inf)
  A -> 1,
]
```
Then, if `C` is moved from position 2 to position 1 (so we have `[B, C, A, D]`):
```js
// -1 is not applicable since this is a backwards move
localChanges = [
  B -> 0,
  D -> 2,
  A -> 1,
  C -> 1,
]
```
Now, if `A` is moved from position 2 to position 3 (so we have `[B, C, D, A]`):
```js
// -1 is not applied anywhere since nothing is in the range (2, +inf) except 3
localChanges = [
  B -> 0,
  D -> 2,
  C -> 1,
  A -> 3,
]
```

ClaytonPassmore:
So I started typing out a long example about how my method was correct but then it totally didn't work haha :-1: 

I buy this. Can we close #184 or are there still issues there?

mattyu007:
Do you think I typed my example up in one pass? 😉 Hopefully third time's the charm. I didn't run into #184 again, but I haven't tested it very rigorously. Ideally if the client side algorithm is correct, then the server shouldn't need any changes.

ClaytonPassmore:
Thanks for making me feel better :laughing: 

Yeah that's what I was thinking.

If we find that there is still an issue, we could maybe modify the server to do like a relative-referral to a card just to get around the whole add and edit fiasco. So for example, we could add a card and if we edit it in the same list of actions, we could add a `refer_relative: index` key-value pair to the edit action that allows us to refer to the index of the action that added the card, and use the uuid from that action. It would be a little contrived but its a possible work-around.

ClaytonPassmore (approved):

ClaytonPassmore:
Nice job! This was a lot of work! Don't forget to rebase and squash :wink:

mattyu007:
Sooooo many crazy merge conflicts

mattyu007:
@ClaytonPassmore Maybe, but right now we would collapse those two into a single `add` action, so there's nothing to refer to. Similarly, multiple moves are always collapsed down into a single action, so we'd need to rethink howe we aggregate local changes.

ClaytonPassmore:
@mattyu007 yeah you're right, basically we wouldn't collapse down add-edits or edit-edits (and maybe more?). But this is gonna work, right? :stuck_out_tongue_winking_eye:

mattyu007:
@ClaytonPassmore Mmmaybe. And also maybe. 😉

PR #180, opened 2017-03-25
Add support for showing network activity indicator on iOS in CueApi

### LINKED ISSUE #176: Add support for network activity indicator in status bar on iOS
This spinny thing in the status bar to indicate network activity:
simulator screen shot mar 24 2017 8 07 25 pm

Update `CueApi` to set the visibility of this indicator based on whether a request is currently processing.

mattyu007:
iOS only, closes #176.

Adds support for the network activity indicator in the status bar for when there are pending `CueApi` calls. Not so useful on speedy networks, but great for if the network is struggling or if we add background sync in the future where we can't show loading indicators in the foreground.

ezgif-3-0399feb644

### DIFF
--- a/Cue/app/api/CueApi.js
+++ b/Cue/app/api/CueApi.js
@@ -1,11 +1,24 @@
 // @flow
 'use strict'
 
+import { StatusBar } from 'react-native'
+
 import { serverURL } from '../env';
 
 class CueApi {
 	static userId  = null;
 	static accessToken = null;
+	static requestsPending = 0;
+
+	static setNetworkActivityIndicatorVisible(value: boolean) {
+		if (value) {
+			this.requestsPending++
+		} else {
+			this.requestsPending--
+		}
+
+		StatusBar.setNetworkActivityIndicatorVisible(this.requestsPending > 0)
+	}
 
 	static setAuthHeader(userId: ?string, accessToken: ?string) {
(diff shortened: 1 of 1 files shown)


Cue/app/api/CueApi.js
```diff
@@ -21,33 +34,36 @@ class CueApi {
...
 		.then(text => {
+			this.setNetworkActivityIndicatorVisible(false)
 			try {
 				return JSON.parse(text)
 			} catch (e) {
 				return {}
 			}
 		})
 		.catch(e => {
+			this.setNetworkActivityIndicatorVisible(false)
```
ClaytonPassmore, line 54: I think it's possible to do a double decrement if the JSON parse fails

ClaytonPassmore (commented):
One minor comment

Cue/app/api/CueApi.js
mattyu007: If JSON parse fails, then we just return an empty object, and the catch should never be reached.

ClaytonPassmore: WOW this is a sign I have been looking at too much code. Not sure how I missed that `try`-`catch` block :confused:

ClaytonPassmore (approved):
#iosMasterRace

PR #185, opened 2017-03-26
Allow for overwriting of deleted flag

### LINKED ISSUE #182: Overwriting a deleted deck doesn't reset deleted flag
If a deck is deleted on the server and you use the overwrite endpoint, the deleted flag will stay at 1.

Expected: The deleted flag should be set to 0 (false)

image

henryc132:
Closes #182 

Allows for overwriting the deletion flag of decks.

Overwriting a deleted deck:
- Sets deletion flag to false
- Adds deck back to user library

### DIFF
--- a/Server/App/api/deck.py
+++ b/Server/App/api/deck.py
@@ -11,6 +11,7 @@ from model.card import Card
 from model.deck import Deck
 from policy.card import CardPolicy
 from policy.deck import DeckPolicy
+from model.library import Library
 from utils.base_handler import BaseHandler
 from utils.wrappers import authorize_request_and_create_db_connector, require_params, \
     extract_user_id, optional_params, optional_query_string_params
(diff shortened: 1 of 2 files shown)


Tyler-Brenneman:
How do you overwrite with a deleted deck?

Looks like you made the deleted flag a required param. Could you update the api spec aswell?

henryc132:
Oh wasn't sure how your conflict logic handled deleted decks, just assumed that you just send the deck to the overwrite point with the deleted flag set to true or something. I guess that part is not needed?

Tyler-Brenneman:
I call the delete endpoint if the localChange is a delete, since we dont have a copy of the local deck anymore if its deleted, just its localchange

henryc132:
Alright, fixed

Server/App/api/deck.py
```diff
@@ -158,11 +159,18 @@ def post(self, name, public, tags, cards, device, user_id, connector, uuid):
                 raise TypeError
             if not isinstance(tags, list):
                 raise TypeError
-
+
```
ClaytonPassmore: Extra whitespace

Server/App/api/deck.py
```diff
@@ -11,6 +11,7 @@
 from model.deck import Deck
 from policy.card import CardPolicy
 from policy.deck import DeckPolicy
+from model.library import Library
```
ClaytonPassmore: Can you put this below  the deck model import? Just to keep things in alphabetical order

ClaytonPassmore (approved):
Looks good! Just two minor things

PR #189, opened 2017-03-26
Only call to update status bar network activity indicator if on iOS

mattyu007:
React Native issues a warning every time you call `StatusBar.setNetworkActivityIndicatorVisible` on Android, which is endlessly annoying.

screen shot 2017-03-26 at 6 14 52 am


ClaytonPassmore (approved):
:ship:

PR #192, opened 2017-03-26
Avoid doubling up TableRow borders on iOS

### LINKED ISSUE #183: Change TableRow borders on iOS so that adjacent rows don't double up on the border
Right now on iOS, each `TableRow` will display borders on both the top and bottom sides, which means that the border between adjacent rows is twice as wide and looks darker than the borders at the top and bottom:

screen shot 2017-03-25 at 6 17 36 pm

It's not so obvious in the screenshot, but it's much more visible on a physical device. We could use `TableHeader` to apply the first border instead, and then have each `TableRow` just draw a border along the bottom.

mattyu007:
iOS only, closes #183.

Minor UI fix, avoids doubling up the border between adjacent `TableRow` elements on iOS by only drawing a bottom border and using `TableHeader` to draw the "top" border of the first row. Right now we don't have any groups of rows without headers, but if we ever do, there's a `showTopBorder` prop on `TableRow` which can be applied to the first row as necessary.

Does nothing on Android.

## Screenshots
Before / After
ezgif-1-d9f3473a5a

### DIFF
--- a/Cue/app/common/TableHeader.js
+++ b/Cue/app/common/TableHeader.js
@@ -1,11 +1,15 @@
 // @flow
 
 import React from 'react'
-import { View, Text, Platform } from 'react-native'
+import { View, Text, StyleSheet, Platform } from 'react-native'
 
 import CueColors from './CueColors'
 
 const iosStyles = {
+  container: {
+    borderBottomColor: CueColors.lightGrey,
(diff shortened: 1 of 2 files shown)


ClaytonPassmore (approved):

ClaytonPassmore:
shweeetttt

PR #196, opened 2017-03-26
Pull version from build config, bump build number to 4

### LINKED ISSUE #195: Pull version info from build manifest
In prepration for distribution, we should grab the version and build number from the app manifest on Android and the Info.plist from iOS when we display it in the app instead of hard-coding a string.

mattyu007:
<sup>High-key I'm really upset that I didn't find out about JS backtick string templates until now.</sup>

Closes #195.

## Summary
This pull request:

1. Updates `AccountHome` and `MenuCreditsItem` to pull the version info from the build configuration
2. Bumps the build number to 4 since the current build number on TestFlight is 3.

## Caveats
Unfortunately we have to update the build number for both iOS (in `Info.plist`) and Android (in `settings.gradle`) manually, but *c'est la vie* unless we actually create some centralized build system, which.. ain't nobody got time fo' that.

## Screenshots
screen shot 2017-03-26 at 6 20 29 pm

### DIFF
--- a/Cue/android/app/build.gradle
+++ b/Cue/android/app/build.gradle
@@ -90,7 +90,7 @@ android {
         applicationId "com.cue"
         minSdkVersion 16
         targetSdkVersion 22
-        versionCode 1
+        versionCode 4
         versionName "1.0"
         ndk {
             abiFilters "armeabi-v7a", "x86"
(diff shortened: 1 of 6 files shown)


ClaytonPassmore (approved):
Sweet!

PR #203, opened 2017-03-27
Normalize share codes before lookup

### LINKED ISSUE #202: Strip spacing and normalize casing before matching share codes
On the client, we display share codes as two sets of four digits to improve readability, but the API expects that the share code is sent as all caps with no spaces in between. Normalize the share code on the client before doing the search.

mattyu007:
Closes #202.

Normalize share codes before lookup by converting to uppercase and stripping all non-alphanumeric characters.

### DIFF
--- a/Cue/app/tabs/library/LibraryHome.js
+++ b/Cue/app/tabs/library/LibraryHome.js
@@ -142,7 +142,7 @@ class LibraryHome extends React.Component {
   _onPressLookUpDeckByShareCode = () => {
     CuePrompt.prompt(
       Platform.OS === 'android' ? 'Look up share code' : 'Look Up Share Code',
-      '',
+      'Spaces and capitalization aren’t important.',
       [
         {text: 'Cancel', style: 'cancel'},
         {text: 'Look Up', onPress: this._onEnterShareCode},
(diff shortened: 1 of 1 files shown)


Cue/app/tabs/library/LibraryHome.js
```diff
@@ -142,18 +142,25 @@ class LibraryHome extends React.Component {
   _onPressLookUpDeckByShareCode = () => {
     CuePrompt.prompt(
       Platform.OS === 'android' ? 'Look up share code' : 'Look Up Share Code',
-      '',
+      'Spaces and capitalization aren’t important.',
```
ClaytonPassmore, line 145: I don't know if we need to include this

ClaytonPassmore (approved):
Looks good! Just not sure if we need the message in the prompt

Cue/app/tabs/library/LibraryHome.js
mattyu007: I don't think it's 100% necessary, but it's extra assurance for users. Kind of like the reverse of how so many password fields still tell you passwords are "case-sensitive".

PR #207, opened 2017-03-27
Expose HTTP status code in error messages

### LINKED ISSUE #206: Expose HTTP status codes in error messages
Right now most of our error messages simply say "Check your Internet connection and try again," which may not be helpful if there's an error.

mattyu007:
Closes #206.

In cases where we have connectivity but receive an error response from the server, we should expose the status code and stop telling people to check their Internet connection.

This also makes it easier to diagnose issues when the app is built in release mode and doesn't expose any React UI for looking deeper into errors.

screen shot 2017-03-27 at 3 33 49 pm
screen shot 2017-03-27 at 3 33 57 pm
screen shot 2017-03-27 at 3 34 08 pm
screen shot 2017-03-27 at 3 34 16 pm
screen shot 2017-03-27 at 3 35 36 pm
screen shot 2017-03-27 at 3 35 55 pm

### DIFF
--- a/Cue/app/api/CueApi.js
+++ b/Cue/app/api/CueApi.js
@@ -54,7 +54,10 @@ class CueApi {
 		})
 		.catch(e => {
 			this.setNetworkActivityIndicatorVisible(false)
-			throw (e)
+			if (!e.recoveryMessage) {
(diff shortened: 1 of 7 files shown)


ClaytonPassmore (approved):
Nice! That will be helpful

PR #209, opened 2017-03-27
Avoid calling API twice for simple deck creates

### LINKED ISSUE #208: Avoid calling Create Deck and Edit Deck API for simple deck creates
Right now, when a deck is created, we call the Create Deck API first, and then call the Edit Deck API to sync card changes and other deck properties.

For simple deck adds (i.e., only adding cards and leaving the deck as private), we can just pass the cards to the Create Deck API and avoid calling the Edit Deck API.

mattyu007:
Closes #208.

The Create Deck API handles deck name, tags, and list of cards. Don't call the Edit Deck API during deck create unless necessary.

Private deck before:
screen shot 2017-03-27 at 3 59 03 pm

Private deck after:
screen shot 2017-03-27 at 4 05 08 pm

Public deck still calls edit deck API:
screen shot 2017-03-27 at 4 14 17 pm

### DIFF
--- a/Cue/app/actions/library.js
+++ b/Cue/app/actions/library.js
@@ -102,7 +102,7 @@ function syncLibrary(localChanges): ThunkAction {
         .catch(e => {
           if (e.response && e.response.status === 409)  // Sync conflict
             failedSyncs.push(change)
-          else // Network error
+          else // Network or server error
             throw e
         }))
     })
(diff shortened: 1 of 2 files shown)


Tyler-Brenneman:
The add deck API only accepts front,back of cards. Couldn't this miss a card flag to the new deck?

mattyu007:
Hmm, good point. I'll take another look to see if this change is actually worth it then, since "creating a deck with no cards" is kind of a rare use case.

PR #210, opened 2017-03-27
Use utf8mb4 and migrate db to utf8mb4

### LINKED ISSUE #204: Server chokes on Unicode characters
Ran into this while creating a Real Deck™. Can't add a card with Korean characters:

screen shot 2017-03-27 at 2 45 36 pm

ClaytonPassmore:
- Have to turn off foreign key checks during migration while you switch character set
- This means that we have the same amount of bytes available in these columns but potentially less characters since certain unicode characters can take up to 4 bytes

Closes #204

### DIFF
--- a/Server/SQL/Migrations/2017-03-27_15-30.sql
+++ b/Server/SQL/Migrations/2017-03-27_15-30.sql
@@ -1,0 +1,46 @@
+USE Cue;
+START TRANSACTION;
+-- BEGIN MIGRATION --
+
+SET foreign_key_checks = 0;
+
+ALTER DATABASE Cue CHARACTER SET = utf8mb4 COLLATE = utf8mb4_unicode_ci;
+
+ALTER TABLE User CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
+ALTER TABLE Deck CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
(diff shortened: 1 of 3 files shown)


mattyu007:
Just to confirm, we're leaving `schema.sql` as-is, and performing the migrations every time we need to initialize a new server DB?

ClaytonPassmore:
@mattyu007 you got it!

ClaytonPassmore:
If you want to apply a migration you just read in the script from the command line util. If you're starting from scratch, you can do `make schema` which will run `schema.sql` and apply all of the migration files in order.

Server/SQL/Migrations/2017-03-27_15-30.sql
```diff
@@ -0,0 +1,46 @@
+USE Cue;
+START TRANSACTION;
+-- BEGIN MIGRATION --
+
+SET foreign_key_checks = 0;
+
+ALTER DATABASE Cue CHARACTER SET = utf8mb4 COLLATE = utf8mb4_unicode_ci;
```
mattyu007, line 7: These equal signs are optional, yes?

mattyu007 (approved):
Seems fine to me.

Server/SQL/Migrations/2017-03-27_15-30.sql
ClaytonPassmore: Yup, optional

Issue #205, opened 2017-03-27
404 when trying to create a share code for a deck which hasn't been synced yet


mattyu007:
Not sure what the right solution here is. The user can enter the flow to create a share code on a new deck which the server doesn't know about yet.

ClaytonPassmore:
Server side stuff fixed with #217

Tyler-Brenneman:
We would have to sync the deck first. Maybe always sync the deck when going to the share options screen?

PR #211, opened 2017-03-28
Automatically trigger sync

### LINKED ISSUE #28: Add support for automatically triggering library sync
Currently the app only supports manually triggering sync by pulling down on the Library home screen. We need to add support for automatically syncing the library at appropriate times, including:
1. When the app is launched
2. When the device transitions from offline to online
3. Potentially when the user returns to the Library home screen by popping the Navigator stack?

---------
## Previous issue
The local data set should be refreshed when the main Library screen comes to the foreground. That is:

1. when the Library screen is shown for the first time (after login or when the app is started)
2. when in-tab navigation pops enough scenes to returns to the Library screen
3. when the Library tab is selected in the tab bar / drawer menu

**1** and **2** can be captured by hooking into the `Component` lifecycle and into the `Navigator`'s navigation context in `LibraryHome.js`:

```js
_navigationEventListenerToken: any

_navigationEventListener(event) {
  // The top-level route is just {}
  if (Object.keys(event.data.route).length === 0) {
    // ** REFRESH ** //
  }
}

componentDidMount() {
  this._navigationEventListenerToken =
    this.props.navigator.navigationContext.addListener('didfocus', this._navigationEventListener)
}

componentWillReceiveProps(nextProps: Props) {
  if (nextProps.navigator !== this.props.navigator) {
    this._navigationEventListenerToken.remove()
    this._navigationEventListenerToken =
      nextProps.navigator.navigationContext.addListener('didfocus', this._navigationEventListener)
  }
}

componentWillUnmount() {
  this._navigationEventListenerToken.remove()
}
```

**3** is TBD, but perhaps we can hook into the `tabs` actions.

mattyu007:
# Please test this on a physical Android device.
My emulator seems to really like [ignoring our error handlers](https://cloud.githubusercontent.com/assets/13400887/24382531/d0f89d5a-1324-11e7-9505-6afbad26467e.png), so I'd like to test this on a Real Android Device™.

Note that the network reachability check on Android is much more naïve than the check on iOS, in that it only checks if you have either an active cell data or WiFi connection, rather than actually testing if the network is usable. If you're testing this in an emulator, you'll need to "disconnect" the device by going into the emulator settings and setting `Cellular > Data status` to `Unregistered (off)`.

## Summary
Closes #28 **but not #205**. This pull request:
1. Adds support for triggering sync based on network availability.
2. Adds a new "Offline Mode" banner under the header in `LibraryHome` to indicate when a device has no network connectivity.
3. Adds support for triggering sync based on whether there are local changes when the Navigator pops to `LibraryHome`.

## Important Caveats
* **In this implementation, `LibraryHome` owns the listener for network availability changes and owns automatically triggering sync.**
    * This works fine on iOS since the component doesn't unmount when switching to a different tab.
    * On Android, even though the component unmounts when the tab is switched, I'm inclined to say that this is OK since in most use cases, `LibraryHome` will be mounted.
    * This does make the component kind of large though.

## Implementation Details
* `LibraryHome` has gained three new pieces of state:
    1. `connected: ?boolean` - whether the network is available
    2. `lastSyncTime: ?Date` - the last time a call to the API succeeded
    3. `needsSync: boolean` - set if a sync is triggered but `LibraryHome` is not in the foreground
* `LibraryHome` has gained two new listeners, each responsible for listening to different state changes:
    1. `_onNavigatorEvent` listens for Navigator events. When `LibraryHome` enters the foreground, checks if `needsSync` is set or if `localChanges` is non-empty. If yes and the network is available, then a sync is triggered.
    2. `_onNetworkIsConnectedChanged` listens for network connectivity changes. If the network becomes available and `LibraryHome` is in the foreground, then a sync is triggered. Otherwise if `LibraryHome` is in the background, then `needsSync` is set to true.

## Screenshots
Offline mode banner:
screen shot 2017-03-27 at 7 56 40 pm

Message when the banner is tapped:
screen shot 2017-03-27 at 7 56 58 pm

Note that the list scrolls *under* the banner:
screen shot 2017-03-27 at 7 58 43 pm

### DIFF
--- a/Cue/app/tabs/library/LibraryHome.js
+++ b/Cue/app/tabs/library/LibraryHome.js
@@ -73,7 +73,7 @@ type Props = {
   onCopyDeck: (deck: Deck) => any,
 }
 
-const NETWORK_SYNC_TRIGGER_THROTTLE_MS = 30000
+const AUTO_SYNC_TRIGGER_THROTTLE_MS = 30000
 
 class LibraryHome extends React.Component {
   props: Props
@@ -156,6 +156,27 @@ class LibraryHome extends React.Component {
     return Object.keys(currentRoute).length === 0
   }
 
+  _shouldThrottleAutoSync = () => {
+    // Throttle calls to _refresh to avoid calling _refresh because:
+    //  1. We can sometimes receive multiple callbacks with [true, false, true]
+    //     in quick succession when the network is reconnecting.
+    //  2. If the SyncConflicts view is dismissed, LibraryHome will re-enter
+    //     the foreground, which can trigger another sync, which will re-push
+    //     SyncConflicts back onto the Navigator stack.
+
+    let msSinceLastSync = this.state.lastSyncTime
+      ? new Date() - this.state.lastSyncTime
+      : Infinity
+    let shouldThrottle = msSinceLastSync < AUTO_SYNC_TRIGGER_THROTTLE_MS
+
+    if (shouldThrottle) {
+      console.warn('_shouldThrottleAutoSync: Auto sync is being throttled (last sync was '
+        + `${msSinceLastSync.toString()} ms ago)`)
+    }
+
+    return shouldThrottle
+  }
+
   _onNavigatorEvent = () => {
     if (this._isLibraryHomeInForeground()) {
       console.info('_onNavigatorEvent: Navigator is at LibraryHome')
@@ -163,11 +184,17 @@ class LibraryHome extends React.Component {
       // Trigger a sync if needsSync is set or if there are any pending localChanges
       if (this.state.needsSync || this.props.localChanges.length) {
         if (this.state.connected) {
-          console.info('_onNavigatorEvent: Sync is pending and network is available; refreshing now '
-            + `(needsSync: ${this.state.needsSync.toString()}, `
-            + `localChanges.length: ${this.props.localChanges.length.toString()})`)
+          if (!this._shouldThrottleAutoSync()) {
+            console.info('_onNavigatorEvent: Sync is pending and network is available; refreshing now '
+              + `(needsSync: ${this.state.needsSync.toString()}, `
+              + `localChanges.length: ${this.props.localChanges.length.toString()})`)
 
-          this._refresh()
+            this._refresh()
+          } else {
+            console.info('_onNavigatorEvent: Sync is pending and network is available but sync is being throttled; doing nothing '
+              + `(needsSync: ${this.state.needsSync.toString()}, `
+              + `localChanges.length: ${this.props.localChanges.length.toString()})`)
+          }
         } else {
           console.info('_onNavigatorEvent: Sync is pending but network is not available; doing nothing '
             + `(needsSync: ${this.state.needsSync.toString()}, `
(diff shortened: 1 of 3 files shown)


Cue/app/tabs/library/LibraryHome.js
```diff
@@ -99,18 +133,104 @@ class LibraryHome extends React.Component {
...
+      let msSinceLastSync = this.state.lastSyncTime
+        ? new Date() - this.state.lastSyncTime
+        : Infinity
+      let shouldThrottle = msSinceLastSync < NETWORK_SYNC_TRIGGER_THROTTLE_MS
+      let shouldSync = isConnected && !shouldThrottle
+
+      if (shouldSync) {
+        console.info('_onNetworkIsConnectedChanged: Requesting sync due to network status change')
+
+        // The top-level route is just {}
```
ClaytonPassmore: Can drop this comment.

ClaytonPassmore (approved):
Awesome! This looks great!

Cue/app/tabs/library/LibraryHome.js
mattyu007: Will do.

mattyu007:
Actually there's a bug right now where the Resolve Conflicts view will keep coming back after being cancelled because `LibraryHome` enters the foreground after `SyncConflicts` pops off the Navigator stack, triggering another sync.

ClaytonPassmore:
@mattyu007 good catch.

Cue/app/tabs/library/LibraryHome.js
```diff
@@ -156,18 +156,45 @@ class LibraryHome extends React.Component {
...
     if (this._isLibraryHomeInForeground()) {
       console.info('_onNavigatorEvent: Navigator is at LibraryHome')
 
       // Trigger a sync if needsSync is set or if there are any pending localChanges
       if (this.state.needsSync || this.props.localChanges.length) {
         if (this.state.connected) {
-          console.info('_onNavigatorEvent: Sync is pending and network is available; refreshing now '
-            + `(needsSync: ${this.state.needsSync.toString()}, `
-            + `localChanges.length: ${this.props.localChanges.length.toString()})`)
+          if (!this._shouldThrottleAutoSync()) {
```
mattyu007: I wish I could collapse this crazy pyramid but this seems like the easiest way to get good logs

mattyu007:
@ClaytonPassmore Ok, we now throttle all auto-triggered sync, so the `SyncConflicts` loop should be mostly resolved, unless someone takes more than 30 seconds to decide that they want to resolve conflicts later.

ClaytonPassmore:
@mattyu007 seems like a reasonable solution. Should we get @lonebutterfish to run it on his Android before merging?

mattyu007:
That would be great! @lonebutterfish

Cue/app/tabs/library/LibraryHome.js
```diff
@@ -99,18 +133,121 @@ class LibraryHome extends React.Component {
...
   _refresh = () => {
     if (!this.state.refreshing) {
       this.setState({refreshing: true})
       this.props.onSyncLibrary(this.props.localChanges).then(failedSyncs =>{
         if (failedSyncs && failedSyncs.length) {
-          this.setState({refreshing: false})
+          this.setState({refreshing: false, lastSyncTime: new Date(), needsSync: false})
           this.props.navigator.push({failedSyncs})
         } else {
           this.props.onLoadLibrary().then(response => {
```
Tyler-Brenneman: When I have no connection the error gets thrown here it is not being caught by the catch at line 254.

This is causing the infinite spinning bug to occur for me and no alert

Tyler-Brenneman (commented):
No idea why this is happening

mattyu007:
ALRIGHT. Fixed the issue with the `onLoadLibrary` promise rejection not being caught. Finally!

Tyler-Brenneman (approved):

PR #213, opened 2017-03-28
Don't display placeholder on iOS in CuePrompt

### LINKED ISSUE #212: Don't display placeholders in prompt dialogs on iOS
On iOS, the placeholder is filled in as the default value, which the user has to clear before they can enter what they want to. We should defer the job of the placeholder elsewhere.

mattyu007:
Closes #212.

Since the placeholder is actually filled in as the default value by `AlertIOS.prompt`, don't display placeholders on iOS to avoid forcing the users to clear the text box before allowing them to enter what they want.

screen shot 2017-03-27 at 9 22 46 pm

### DIFF
--- a/Cue/app/tabs/library/LibraryHome.js
+++ b/Cue/app/tabs/library/LibraryHome.js
@@ -178,7 +178,7 @@ class LibraryHome extends React.Component {
     const MAX_LENGTH = 255
     CuePrompt.prompt(
       Platform.OS === 'android' ? 'Create new deck' : 'Create New Deck',
-      'Enter a name for your new deck.',
+      'Enter a name for your deck.',
       [
         {text: 'Cancel', style: 'cancel'},
         {text: 'Create', onPress: (deckName) => {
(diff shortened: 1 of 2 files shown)


Cue/app/tabs/library/LibraryHome.js
```diff
@@ -178,7 +178,7 @@ class LibraryHome extends React.Component {
     const MAX_LENGTH = 255
     CuePrompt.prompt(
       Platform.OS === 'android' ? 'Create new deck' : 'Create New Deck',
-      '',
+      'Enter a name for your new deck.',
```
ClaytonPassmore: Nit: Saying "new deck" in both the title and the message seems repetitive. What if we just say "Enter a name for your deck"?

ClaytonPassmore (approved):
Looks good! One minor comment.

Cue/app/tabs/library/LibraryHome.js
mattyu007: Good point.

PR #216, opened 2017-03-28
Change naming of copied deck

### LINKED ISSUE #215: Deck Copy Behaviour
When copying a public deck, name the deck "Copy of <original deck name" so that it's less confusing for the users.

ClaytonPassmore:
Closes #215 

As you would expect:
deckcopy


mattyu007 (approved):
Short and sweet.

PR #217, opened 2017-03-28
None check on deck share code handler

ClaytonPassmore:
I missed the `None` check.

Return 404 now rather than 500 (oops).

Affects (but does not close) #205.


henryc132 (approved):

henryc132:
Kind of feels weird that DeckPolicy.owned_by is used to get the initial deck

ClaytonPassmore:
@henryc132 its because you can't generate a share code for a deck unless you own it.

PR #221, opened 2017-03-28
Allow number of columns in LibraryListView to vary

### LINKED ISSUE #220: Allow number of columns in LibraryListView to vary
Further to #219, we should allow the number of deck thumbnails per column in `LibraryListView` to vary based on the viewport width. Currently we are restricted to two columns, which works great in portrait, but looks really goofy in landscape and on iPad.

iPhone 6s Plus in landscape:
img_1842

mattyu007:
Closes #220. 

## Summary
This pull request:
1. Allows the number of columns in `LibraryListView` to vary by dynamically calculating the width of `DeckThumbnail` based on the current window width
2. Forces the list to re-render on orientation change in `LibraryListView`, `DiscoverHome`, and `SearchListView` to allow `DeckThumbnail` to re-render using the updated window width

Building on #219 which added iPad support, this makes `LibraryListView` more comfortable to use in landscape and on larger tablet-sized devices.

## Screenshots
Portrait still looks like it used to:
screen shot 2017-03-28 at 2 31 51 am

Landscape now has 3 columns instead of 2:
screen shot 2017-03-28 at 2 32 25 am

iPad now looks great in both orientations:
img_0349
img_0350

Also works on Nexus 10 (don't ask me why the shadows from the `elevation` are like that, I have no idea):
screen shot 2017-03-28 at 2 51 44 am

### DIFF
--- a/Cue/app/common/DeckThumbnail.js
+++ b/Cue/app/common/DeckThumbnail.js
@@ -88,13 +88,32 @@ export default class DeckThumbnail extends React.Component {
     onPressDelete?: () => void,
   }
 
+  _getWidth = () => {
+    const paddingOneSide = 16
+    const minWidth = 160
+    const windowWidth = Dimensions.get('window').width
+
+    // With n columns, we have n + 1 instances of 16 pt horizontal margins,
+    // so subtract one extra 16 pt from the window width to get the number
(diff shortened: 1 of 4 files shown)


ClaytonPassmore (approved):
Nice!

PR #223, opened 2017-03-28
Relax restrictions on characters allowed in tags

### LINKED ISSUE #222: Be less zealous when sanitizing tags
Right now we force tags to fit into the regex `/^[a-z ]*$/i` which is extremely restrictive and excludes common cases such as using course codes as tags. We should allow numerics and normalize diacritics down to regular ASCII characters.

mattyu007:
Closes #222.

## Summary
This pull request:
1. Relaxes the restrictions on tags to allow any character
2. Disallows storing tags which compare equal when diacritics are stripped.

----
# Previously:

## Summary
This pull request:
1. Allows numbers in tags
2. Strips diacritics before filtering out non-alphanumeric characters in tags

@ClaytonPassmore @henryc132 Does have some unforeseen negative effect on full-text search on the server?

ezgif-1-fe10f53248

### DIFF
--- a/Cue/app/tabs/library/TagsTableRow.js
+++ b/Cue/app/tabs/library/TagsTableRow.js
@@ -4,7 +4,6 @@
 
 import React from 'react'
 import { View, Text, TextInput, Platform } from 'react-native'
-import { remove as stripDiacritics } from 'diacritics'
 
 import Chip from '../../common/Chip'
 import CueColors from '../../common/CueColors'
@@ -48,10 +47,9 @@ export default class TagsTableRow extends React.Component {
   _onEndEditing = ({nativeEvent: {text}}) => {
     this.textInputRef.clear()
 
-    let tag = stripDiacritics(text)
+    let tag = text
       .trim()
       .toLowerCase()
-      .replace(new RegExp('[^-a-z0-9 ]', 'g'), '')
       .replace(new RegExp(' +', 'g'), ' ')
 
     if (tag.length > 0) {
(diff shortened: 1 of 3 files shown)


ClaytonPassmore:
I don't see why it would but @henryc132 might have a better idea.

mattyu007:
Actually...  why are we so restrictive about the contents of tags anyway, isn't the only real requirement that they don't contain any commas? Is it just to make search easier?

ClaytonPassmore:
@mattyu007 as far as I know, yeah, we don't really need to be so restrictive.. its just for searchability

Cue/app/tabs/library/TagsTableRow.js
```diff
@@ -47,8 +47,9 @@ export default class TagsTableRow extends React.Component {
   _onEndEditing = ({nativeEvent: {text}}) => {
     this.textInputRef.clear()
 
-    let tag = text.trim().toLowerCase()
-      .replace(new RegExp('[^-a-z ]', 'g'), '')
+    let tag = text
+      .trim()
+      .toLowerCase()
```
ClaytonPassmore, line 52: Does this work even with non-lowercase-able characters?

ClaytonPassmore (commented):
Small question

Cue/app/tabs/library/TagsTableRow.js
mattyu007: Indeed. I didn't see any adverse effects trying it in the simulator with Korean or other Japanese characters either.

screen shot 2017-03-28 at 9 17 23 pm

ClaytonPassmore (approved):
Nice job!

PR #225, opened 2017-03-28
Turn on "Flagged cards only" if deck view was filtering

ClaytonPassmore:
We originally had this behavior but somewhere along the line it broke (maybe it was me? IDK works now :stuck_out_tongue_closed_eyes: )


mattyu007 (approved):
Aah this might have been me messing up a merge conflict after renaming the state variable.

PR #227, opened 2017-03-28
(Mostly) properly centre indicators when swiping in Play Mode

mattyu007:
Previously the calculation for how far off the screen to move the card didn't account for the 0.4:1 damping, width, and horizontal padding of the indicators, so on wider viewports, the icons would be visibly misaligned when the card was released.


ClaytonPassmore (approved):
I believe you

PR #230, opened 2017-03-29
Sort decks in Library alphabetically and update build config

mattyu007:
Cue ~v1.0.1~ v1.0! Finally.

## Summary
This pull request:
- Sort decks in LibraryListView
- Bump build version to 7
- Set applicationId for Android app to match that of iOS

### DIFF
--- a/Cue/app/tabs/library/LibraryListView.js
+++ b/Cue/app/tabs/library/LibraryListView.js
@@ -61,6 +61,15 @@ class LibraryListView extends React.Component {
   }
 
   _categorizeDecks(decks: Array<Deck>) {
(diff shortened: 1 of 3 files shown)


Cue/app/tabs/library/LibraryListView.js
```diff
@@ -61,6 +61,15 @@ class LibraryListView extends React.Component {
   }
 
   _categorizeDecks(decks: Array<Deck>) {
+    decks.sort((a: Deck, b: Deck) => {
```
ClaytonPassmore: Should we be slicing so we don't modify in place?

ClaytonPassmore (approved):
:ship:

